PR #2, opened 2017-01-12
Add initial unit tests for arcade drive control

shadaj:
Adds basic control tests for open-loop control and feedforward only control (basically open-loop).

Shows the importance of unit testing because this found a bug with our PIDF control method (switched sensor signal with target for feedforward)!


ghost (approved):
LGTM

PR #3, opened 2017-01-14
Upgrade FRC module to use 2017 WPILib

shadaj:
Had to add some overriding classes to prevent JNI libs from trying to load.


LordDarkula (approved):
lgtm

PR #4, opened 2017-01-14
Add model module with support for simulating state space models

shadaj:


model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
```diff
@@ -0,0 +1,41 @@
+package com.lynbrookrobotics.potassium.model.examples
+
+import breeze.linalg.{DenseMatrix, DenseVector}
+import breeze.plot._
+import com.lynbrookrobotics.potassium.model.StateSpace
+
+object FallingMass extends App {
+  val NumIterations = 1000
+
+  val times = new Array[Double](NumIterations)
```
PhilipAxelrod: nit, align equal signs

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
```diff
@@ -0,0 +1,41 @@
...
+  val positions = new Array[Double](NumIterations)
+
+  val Mass = 1 // kg
```
PhilipAxelrod: Why is this not using squant units?

PhilipAxelrod: also, nit: first letter should be lower case for a variable name

shadaj: Good catch! I'll fix that.

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
```diff
@@ -0,0 +1,41 @@
...
+  val GravityForce = Mass * 9.8 // N
```
PhilipAxelrod: Why is this not using squant units?
nit: first letter should be lower case for a variable name

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
```diff
@@ -0,0 +1,41 @@
...
+      (0D, 1D),
+      (0D, 0D)
+    ),
+    DenseMatrix(
+      0,
+      -GravityForce
+    )
+  )
+
+  var state = DenseVector(0D, 0D)
```
PhilipAxelrod, line 33: So the first value represents time, the second position?

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  for (i <- 0 until NumIterations) {
+    times(i) = i * 0.005
```
PhilipAxelrod: Where is .005 coming from?

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
shadaj: It's a constant here, so we use camelcase

model/src/test/scala/com/lynbrookrobotics/potassium/model/examples/FallingMass.scala
shadaj: First is position, second is velocity.

PhilipAxelrod (approved):
lgtm

PR #5, opened 2017-01-14
Remove second type parameter for SignalLike and add alternative PeriodicSignal map with no dt

shadaj:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
```diff
@@ -9,8 +9,8 @@ trait Drive {
   type DriveSignal
   type DriveVelocity
 
-  protected def driveClosedLoop[C[_]](signal: SignalLike[DriveSignal, C]): PeriodicSignal[DriveSignal]
-  protected def driveVelocity[C[_]](velocity: SignalLike[DriveVelocity, C]): PeriodicSignal[DriveSignal]
+  protected def driveClosedLoop(signal: SignalLike[DriveSignal]): PeriodicSignal[DriveSignal]
+  protected def driveVelocity(velocity: SignalLike[DriveVelocity]): PeriodicSignal[DriveSignal]
```
PhilipAxelrod, line 13: What is the difference between these two? One controls velocity, the other motor power?

shadaj: The first takes a target percentage (like what joysticks would have) and runs velocity control on that. The second takes a velocity (in proper velocity units) and controls to that. I really want to clean this up at some point, but have to figure out how to do that without forcing anyone who has non-velocity units signals to first convert them to velocity units before running velocity control.

shadaj:
Went through and switched to using the new `map` wherever possible. It's a lot lot cleaner now!

PhilipAxelrod (approved):
lgtm

PR #6, opened 2017-01-14
Only enable coverage when running unit tests

shadaj:
This prevents publishing libraries with coverage enabled inside them

Discovered when running frc-2017 with @dtjong


ghost (approved):
LGTM

PR #7, opened 2017-01-16
Add drivetrain hardware/props and cleanup controller method names

shadaj:
Because we want to pass around hardware and properties values instead of directly depending on them (for live reloading, unit testing, etc), there are now DrivetrainHardware (for hardware interfaces) and DrivetrainProperties (for configuration) classes. We depend on these using implicit parameters to make method usage clean (subject to change???).

Also changes some controller names to clarify open-loop vs closed-loop vs velocity control.

Still need more cleanup for this, but I think this puts us in a good position to be able to start adding position and other more involved controllers. We'll probably end up landing upon better organization as we write those, anyway.

### DIFF
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
@@ -9,10 +9,18 @@ trait Drive {
   type DriveSignal
   type DriveVelocity
 
-  protected def driveClosedLoop(signal: SignalLike[DriveSignal]): PeriodicSignal[DriveSignal]
-  protected def driveVelocity(velocity: SignalLike[DriveVelocity]): PeriodicSignal[DriveSignal]
+  type DrivetrainHardware
+  type DrivetrainProperties
 
-  protected def defaultController: PeriodicSignal[DriveSignal]
+  /**
+    * Drives with the signal with closed-loop control
+    * @param signal the signal to drive with
+    * @return
+    */
+  protected def driveClosedLoop(signal: SignalLike[DriveSignal]): PeriodicSignal[DriveSignal]
+
+  protected def defaultController(implicit hardware: DrivetrainHardware,
+                                  props: DrivetrainProperties): PeriodicSignal[DriveSignal]
 
   type Drivetrain <: Component[DriveSignal]
 }
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
@@ -1,12 +1,37 @@
 package com.lynbrookrobotics.potassium.commons.drivetrain
 
 import com.lynbrookrobotics.potassium.clock.Clock
-import com.lynbrookrobotics.potassium.control.{PIDF, PIDFConfig, PIDFProperUnitsConfig}
+import com.lynbrookrobotics.potassium.control.{PIDF, PIDFConfig}
 import com.lynbrookrobotics.potassium.units._
 import com.lynbrookrobotics.potassium.{Component, PeriodicSignal, Signal, SignalLike}
-import squants.{Acceleration, Dimensionless, Each, Length, Time, Velocity}
+
 import squants.motion.{MetersPerSecond, MetersPerSecondSquared}
 import squants.space.Meters
+import squants.{Dimensionless, Each, Time, Velocity}
+
+trait TwoSidedDriveHardware extends UnicycleHardware {
+  val leftVelocity: Signal[Velocity]
+  val rightVelocity: Signal[Velocity]
+
+  val forwardVelocity: Signal[Velocity] =
+    leftVelocity.zip(rightVelocity).map(t => (t._1 + t._2) / 2)
+}
+
+trait TwoSidedDriveProperties extends UnicycleProperties {
+  val maxLeftVelocity: Velocity
+  val maxRightVelocity: Velocity
+  val leftControlGains: VelocityGains
+  val rightControlGains: VelocityGains
+
+  val maxForwardVelocity: Velocity = maxLeftVelocity min maxRightVelocity
+
+  val forwardControlGains: VelocityGains = PIDFConfig(
+    Each(0) / MetersPerSecond(1),
+    Each(0) / MetersPerSecondSquared(1),
+    Each(0) / Meters(1),
+    Each(1) / maxForwardVelocity
+  )
+}
 
 /**
   * A drivetrain with two side control (such as a tank drive)
@@ -18,7 +43,8 @@ trait TwoSidedDrive extends UnicycleDrive { self =>
   type DriveSignal = TwoSidedSignal
   type DriveVelocity = TwoSidedVelocity
 
-  type DrivetrainHardware
+  type DrivetrainHardware <: TwoSidedDriveHardware
+  type DrivetrainProperties <: TwoSidedDriveProperties
 
   protected implicit val clock: Clock
   protected val updatePeriod: Time
@@ -32,46 +58,45 @@ trait TwoSidedDrive extends UnicycleDrive { self =>
     )
   }
 
-  protected def expectedVelocity(drive: TwoSidedSignal): TwoSidedVelocity = {
-    TwoSidedVelocity(maxLeftVelocity * drive.left, maxRightVelocity * drive.right)
+  protected def expectedVelocity(drive: TwoSidedSignal)(implicit props: DrivetrainProperties): TwoSidedVelocity = {
+    TwoSidedVelocity(props.maxLeftVelocity * drive.left, props.maxRightVelocity * drive.right)
   }
 
-  protected def driveClosedLoop(signal: SignalLike[TwoSidedSignal]): PeriodicSignal[TwoSidedSignal] =
-    TwoSidedControllers.velocity(signal.map(expectedVelocity))
-
-  protected def driveVelocity(signal: SignalLike[TwoSidedVelocity]): PeriodicSignal[TwoSidedSignal] =
-    TwoSidedControllers.velocity(signal)
-
-  protected val maxLeftVelocity: Velocity
-  protected val maxRightVelocity: Velocity
-
-  protected val leftControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
-  protected val rightControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
-
-  // disable unicycle forward control because we already control that here
-  override protected val forwardControlGains: PIDFConfig[Velocity, Velocity, Acceleration, Length, Dimensionless] = PIDFConfig(
-    Each(0) / MetersPerSecond(1),
-    Each(0) / MetersPerSecondSquared(1),
-    Each(0) / Meters(1),
-    Each(0) / MetersPerSecond(0)
-  )
-
-  protected val leftVelocity: PeriodicSignal[Velocity]
-  protected val rightVelocity: PeriodicSignal[Velocity]
-
-  override protected val forwardVelocity: PeriodicSignal[Velocity] =
-    leftVelocity.zip(rightVelocity).map(t => (t._1 + t._2) / 2)
+  protected def driveClosedLoop(signal: SignalLike[TwoSidedSignal])
+                               (implicit hardware: DrivetrainHardware,
+                                         props: DrivetrainProperties): PeriodicSignal[TwoSidedSignal] =
+    TwoSidedControllers.closedLoopControl(signal)
 
   object TwoSidedControllers {
-    def velocity(target: SignalLike[TwoSidedVelocity]): PeriodicSignal[TwoSidedSignal] = {
-      val leftControl = PIDF.pidf(leftVelocity, target.map(_.left).toPeriodic, leftControlGains)
-      val rightControl = PIDF.pidf(rightVelocity, target.map(_.right).toPeriodic, rightControlGains)
+    def velocityControl(target: SignalLike[TwoSidedVelocity])
+                       (implicit hardware: DrivetrainHardware,
+                                 props: DrivetrainProperties): PeriodicSignal[TwoSidedSignal] = {
+      import hardware._
+      import props._
+
+      val leftControl = PIDF.pidf(
+        leftVelocity.toPeriodic,
+        target.map(_.left).toPeriodic,
+        leftControlGains
+      )
+
+      val rightControl = PIDF.pidf(
+        rightVelocity.toPeriodic,
+        target.map(_.right).toPeriodic,
+        rightControlGains
+      )
 
       leftControl.zip(rightControl).map(s => TwoSidedSignal(s._1, s._2))
     }
+
+    def closedLoopControl(signal: SignalLike[TwoSidedSignal])
+                         (implicit hardware: DrivetrainHardware,
+                                   props: DrivetrainProperties): PeriodicSignal[TwoSidedSignal] = {
+      velocityControl(signal.map(expectedVelocity))
+    }
   }
 
-  class Drivetrain(implicit hardware: DrivetrainHardware) extends Component[TwoSidedSignal](updatePeriod) {
+  class Drivetrain(implicit hardware: DrivetrainHardware, props: DrivetrainProperties) extends Component[TwoSidedSignal](updatePeriod) {
     override def defaultController: PeriodicSignal[TwoSidedSignal] = self.defaultController
 
     override def applySignal(signal: TwoSidedSignal): Unit = {
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
@@ -1,12 +1,30 @@
 package com.lynbrookrobotics.potassium.commons.drivetrain
 
 import com.lynbrookrobotics.potassium.control.{PIDF, PIDFConfig, PIDFProperUnitsConfig}
-import com.lynbrookrobotics.potassium.{PeriodicSignal, Signal, SignalLike}
 import com.lynbrookrobotics.potassium.tasks.ContinuousTask
+import com.lynbrookrobotics.potassium.units.GenericValue._
 import com.lynbrookrobotics.potassium.units._
+import com.lynbrookrobotics.potassium.{PeriodicSignal, Signal, SignalLike}
+
 import squants.motion.AngularVelocity
 import squants.{Acceleration, Dimensionless, Length, Percent, Velocity}
-import GenericValue._
+
+trait UnicycleProperties {
+  val maxForwardVelocity: Velocity
+  val maxTurnVelocity: AngularVelocity
+
+  val forwardControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
+  val turnControlGains: PIDFConfig[AngularVelocity,
+                                   GenericValue[AngularVelocity],
+                                   GenericDerivative[AngularVelocity],
+                                   GenericIntegral[AngularVelocity],
+                                   Dimensionless]
+}
+
+trait UnicycleHardware {
+  val forwardVelocity: Signal[Velocity]
+  val turnVelocity: Signal[AngularVelocity]
+}
 
 /**
   * A drivetrain that has forward-backward and turning control in the unicycle model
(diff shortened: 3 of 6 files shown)


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
```diff
@@ -9,10 +9,18 @@ trait Drive {
...
 
+  type DrivetrainHardware
+  type DrivetrainProperties
+
+  /**
+    * Drives with the signal with closed-loop control
+    * @param signal the signal to drive with
+    * @return
+    */
   protected def driveClosedLoop(signal: SignalLike[DriveSignal]): PeriodicSignal[DriveSignal]
```
PhilipAxelrod, line 20: Shouldn't this be public?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
...
+  val forwardControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
+  val turnControlGains: PIDFConfig[AngularVelocity,
+                                   GenericValue[AngularVelocity],
+                                   GenericDerivative[AngularVelocity],
+                                   GenericIntegral[AngularVelocity],
+                                   Dimensionless]
+}
+
+trait UnicycleHardware {
+  val forwardVelocity: PeriodicSignal[Velocity]
```
PhilipAxelrod: I don't think this would be a periodic signal. Sensors like the gyro and encoders don't need periodic updates to tell the velocity. Same with turnVelocity

PhilipAxelrod: Perhaps this should be a signalLike or just signal instead?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -8,6 +8,28 @@ import squants.{Acceleration, Dimensionless, Each, Length, Time, Velocity}
 import squants.motion.{MetersPerSecond, MetersPerSecondSquared}
 import squants.space.Meters
 
+trait TwoSidedDriveHardware extends UnicycleHardware {
+  val leftVelocity: PeriodicSignal[Velocity]
```
PhilipAxelrod: Not sure this needs to be periodic, as encoders and gyro that measure this don't require keeping a log / constant updating for this feature

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -8,6 +8,28 @@ import squants.{Acceleration, Dimensionless, Each, Length, Time, Velocity}
...
+  val rightVelocity: PeriodicSignal[Velocity]
+
+  val forwardVelocity: PeriodicSignal[Velocity] =
+    leftVelocity.zip(rightVelocity).map(t => (t._1 + t._2) / 2)
+}
+
+trait TwoSidedDriveProperties extends UnicycleProperties {
+  val maxLeftVelocity: Velocity
+  val maxRightVelocity: Velocity
+  val leftControlGains: VelocityGains
```
PhilipAxelrod, line 23: Wait, so what exactly is the purpose of this? For dealing with problems where one side of the drivetrain has more drag than the other? If so, how does that relate to forwardControlGains, if at all?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
```diff
@@ -9,10 +9,18 @@ trait Drive {
   type DriveSignal
   type DriveVelocity
 
+  type DrivetrainHardware
```
PhilipAxelrod, line 12: I don't quite understand why we are having these as a type aliases and not as members

PhilipAxelrod: In the end, DrivetrainHardware and DrivetrainProps will be objects anyways, right?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -8,6 +8,28 @@ import squants.{Acceleration, Dimensionless, Each, Length, Time, Velocity}
...
+  val forwardControlGains: VelocityGains = PIDFConfig(
+    Each(0) / MetersPerSecond(1),
+    Each(0) / MetersPerSecondSquared(1),
+    Each(0) / Meters(1),
+    Each(0) / MetersPerSecond(0)
+  )
+}
+
 /**
   * A drivetrain with two side control (such as a tank drive)
```
PhilipAxelrod, line 37: So these like replacements for components in 2016. Everything that is a Drive is the equivalent of a being a Drivetrain component in 2016?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -32,46 +55,35 @@ trait TwoSidedDrive extends UnicycleDrive { self =>
...
-  protected val rightVelocity: PeriodicSignal[Velocity]
-
-  override protected val forwardVelocity: PeriodicSignal[Velocity] =
-    leftVelocity.zip(rightVelocity).map(t => (t._1 + t._2) / 2)
+  protected def driveClosedLoop(signal: SignalLike[TwoSidedSignal])
+                               (implicit hardware: DrivetrainHardware,
+                                         props: DrivetrainProperties): PeriodicSignal[TwoSidedSignal] =
+    TwoSidedControllers.closedLoopControl(signal)
 
   object TwoSidedControllers {
```
PhilipAxelrod, line 70: This should not be declared inside of another class. It makes it hard to extend, which we will almost certainly be doing

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
 import squants.{Acceleration, Dimensionless, Length, Percent, Velocity}
 import GenericValue._
 
+trait UnicycleProperties {
+  val maxForwardVelocity: Velocity
+  val maxTurnVelocity: AngularVelocity
+
+  val forwardControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
+  val turnControlGains: PIDFConfig[AngularVelocity,
+                                   GenericValue[AngularVelocity],
```
PhilipAxelrod, line 18: So, what is the purpose of this GenericValue class? What is the purpose of GenericIntegral and GenericDerivitive?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
 import squants.{Acceleration, Dimensionless, Length, Percent, Velocity}
 import GenericValue._
 
+trait UnicycleProperties {
+  val maxForwardVelocity: Velocity
+  val maxTurnVelocity: AngularVelocity
+
+  val forwardControlGains: PIDFProperUnitsConfig[Velocity, Acceleration, Length, Dimensionless]
```
PhilipAxelrod, line 16: So I took a look at this, and I don't really see the point of it. Could you please explain why you think this is necessary?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
...
+    */
+  private def parentClosedLoop(unicycle: SignalLike[UnicycleSignal]): PeriodicSignal[DriveSignal] = {
     driveClosedLoop(unicycle.map(convertUnicycleToDrive))
   }
 
   object UnicycleControllers {
-    def forwardOpen(forwardSpeed: Signal[Dimensionless]): PeriodicSignal[DriveSignal] = {
-      toClosedDrive(forwardSpeed.map(f => UnicycleSignal(f, Percent(0))))
+    def forwardOpenLoop(forwardSpeed: Signal[Dimensionless]): PeriodicSignal[DriveSignal] = {
+      parentClosedLoop(forwardSpeed.map(f => UnicycleSignal(f, Percent(0))))
```
PhilipAxelrod, line 66: So where does parentClosedLoop come into play? Shouldn't this just be 
forwardSpeed.map(f => UnicycleSignal(f, Percent(0))) , by the definition of open loop?

PhilipAxelrod: Plus, the way this is now, it's not really open loop.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
...
     }
 
-    def turnOpen(turnSpeed: Signal[Dimensionless]): PeriodicSignal[DriveSignal] = {
-      toClosedDrive(turnSpeed.map(t => UnicycleSignal(Percent(0), t)))
+    def turnOpenLoop(turnSpeed: Signal[Dimensionless]): PeriodicSignal[DriveSignal] = {
+      parentClosedLoop(turnSpeed.map(t => UnicycleSignal(Percent(0), t)))
```
PhilipAxelrod, line 70: Same comment as above

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
...
     }
 
-    def velocity(target: SignalLike[UnicycleVelocity]): PeriodicSignal[UnicycleSignal] = {
+    def velocityControl(target: SignalLike[UnicycleVelocity])
+                       (implicit hardware: DrivetrainHardware,
+                        props: DrivetrainProperties): PeriodicSignal[UnicycleSignal] = {
+      import hardware._, props._
+
       val forwardControl = PIDF.pidf(forwardVelocity, target.map(_.forward).toPeriodic, forwardControlGains)
```
PhilipAxelrod: Not 100% certain why the signal for current forward velocity has to be periodic

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -8,55 +8,84 @@ import squants.motion.AngularVelocity
...
     }
 
-    def expectedVelocity(unicycle: SignalLike[UnicycleSignal]): PeriodicSignal[UnicycleSignal] = {
-      velocity(unicycle.map(s => UnicycleVelocity(
+    def closedLoopControl(unicycle: SignalLike[UnicycleSignal])
+                         (implicit hardware: DrivetrainHardware,
+                          props: DrivetrainProperties): PeriodicSignal[UnicycleSignal] = {
+      import props._
+
+      velocityControl(unicycle.map(s => UnicycleVelocity(
```
PhilipAxelrod, line 99: by taking a UnicycleVelocity as a paramter instead, this would be neater

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -65,11 +94,11 @@ trait UnicycleDrive extends Drive {
...
 
   object unicycleTasks {
-    class ContinuousDrive(forward: Signal[Dimensionless], turn: Signal[Dimensionless])
-                         (implicit drive: Drivetrain) extends ContinuousTask {
+    class ContinuousClosedDrive(forward: Signal[Dimensionless], turn: Signal[Dimensionless])
+                               (implicit drive: Drivetrain) extends ContinuousTask {
       override def onStart(): Unit = {
         val combined = forward.zip(turn).map(t => UnicycleSignal(t._1, t._2))
-        drive.setController(toClosedDrive(combined))
+        drive.setController(parentClosedLoop(combined))
```
PhilipAxelrod, line 112: I don't understand the whole purpose of this parentClosedLoop thing. What is the point, and what does it do?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
shadaj: We want to take hardware and props as parameters in all controllers/tasks, for unit testing purposes, so here we just declare an abstract type for what type of hardware and props the methods should be receiving.

Now that I think of it, we can probably just eliminate these types and take the unicycle/twosided hardware and props directly, and this should work since the final hardware/props case class will extend those types.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
shadaj: Yeah agree, I'll change this (and other hardware signals).

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
shadaj: Yeah, that's correct for taking the two gains. Also generally in case we want to use it in some weird situation where gear ratios are different or something.

Unicycle drivetrains take a props with forward control gains, since unicycles have forward/backward and left/right control. But since we're essentially controlling the forward/backward speed through the left/right gains, we disable forward/backward control.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -8,6 +8,28 @@ import squants.{Acceleration, Dimensionless, Each, Length, Time, Velocity}
...
+    leftVelocity.zip(rightVelocity).map(t => (t._1 + t._2) / 2)
+}
+
+trait TwoSidedDriveProperties extends UnicycleProperties {
+  val maxLeftVelocity: Velocity
+  val maxRightVelocity: Velocity
+  val leftControlGains: VelocityGains
+  val rightControlGains: VelocityGains
+
+  val forwardControlGains: VelocityGains = PIDFConfig(
```
shadaj, line 28: This isn't actually the right way to disable gains, since we should keep the feed forward gain to normal. I'll fix this.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
shadaj: Yup!

shadaj: Actually, kinda. This is more of a package that includes the component, controllers, and tasks. That's why we have objects inside the trait, since we expect to mix in the traits into a package object so you get the feel of having a proper package with components, controllers, tasks, etc. inside.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
shadaj: See above, we want it as an object since the trait will be mixed into a package. So this object will be more like a proper object inside the package instead of an object inside of a class.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: All we're saying is that we want to store the PIDf gains for controlling the forward/backward speed, where the signal type is `Velocity`, the derivative type is `Acceleration`, the integral type is `Length`, and the output type is `Dimensionless`.

We really only take these gains in the case of some weird type of drivetrain (like an actual unicycle!) where we want to control forward/backward speed at the unicycle level. But in most cases this control will be disabled by the drivetrain trait that extends it.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: In Squants, AngularVelocity doesn't have a derivative or integral, but to do PID we want units that represent these values. So we have GenericValue (which wraps another unit), which can be integrated to a GenericIntegral or derived to a GenericDerivative.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: Yea, I'm switching it to a signal.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: That's a really good point, it's open loop at the unicycle level, but we pass the resulting speeds to closed loop at the parent (two-sided drive) level. What do you think would be the right behavior/naming here?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: Here we want to take percentages, convert them to velocities and then pass it to velocityControl, which is exactly what you described with `UnicycleVelocity` as a parameter.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: That is the method to pass drive signals to to run closed loop at the parent (whoever's extending `UnicycleDrive`) level. So here we would be running left-right control at the `TwoSidedDrive` level.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
shadaj: It's something that whoever extends this trait has to implement, only used internally and not as a proper controller.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/Drive.scala
shadaj: Nevermind the second point, because `Drive` needs knowledge of the types for the `defaultController` method.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
PhilipAxelrod: The derivative of angular speed is angular acceleration, and its integral is angle. Why not just use these units instead?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
PhilipAxelrod: I think that we should do 1 of 2 things:
1) remove the parentClosedLoop method, making it truly open loop
or
2) Rename this method to driveClosedLoopAtSpeed or something similar to clarify the purpose

PhilipAxelrod: I think option 2 probably makes more sense, because we really have no need of open loop control in the drivetrain

kunalsheth (approved):
Philip is the god.

shadaj:
Hmm, all the tests pass locally. Going to rerun Travis.

PR #9, opened 2017-01-18
Add more examples for LTI models

RinHa:
Added an example that simulates two forces acting upon an object and added another example that simulates a hanging spring that is dropped from the top.


shadaj (approved):
LGTM 🎉

PR #13, opened 2017-01-24
Added two way propagated signals with file and JSON implementations

RinHa:


config/src/main/scala/com/lynbrookrobotics/potassium/config/TwoWayFile.scala
```diff
@@ -0,0 +1,15 @@
+package com.lynbrookrobotics.potassium.config
+
+import java.io.{File, FileOutputStream, PrintWriter}
+
+class TwoWayFile(file: File) extends TwoWaySignal[String]{
```
shadaj: Nit: space before {

shadaj (changes requested):
Just one nit!

shadaj (approved):
LGTM 🥇

PR #18, opened 2017-01-27
Add trapezoidal profile drivetrain controller.

PhilipAxelrod:
Adds drivetrain controller with the following behaviour. A constant acceleration to a cruising velocity, followed by maintaining a cruising velocity, followed by a constant decceleration, getting to the target position at the target final velocity.

https://app.asana.com/0/219272915179160/250106024468357

### DIFF
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
@@ -49,6 +49,7 @@ trait TwoSidedDrive extends UnicycleDrive { self =>
   protected implicit val clock: Clock
   protected val updatePeriod: Time
 
+  //TODO: add docs
   protected def output(hardware: DrivetrainHardware, signal: TwoSidedSignal): Unit
 
   protected def convertUnicycleToDrive(uni: UnicycleSignal): TwoSidedSignal = {
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
@@ -26,13 +26,13 @@ trait UnicycleHardware {
   val turnVelocity: Signal[AngularVelocity]
 }
 
+case class UnicycleSignal(forward: Dimensionless, turn: Dimensionless)
+case class UnicycleVelocity(forward: Velocity, turn: AngularVelocity)
+
 /**
   * A drivetrain that has forward-backward and turning control in the unicycle model
   */
 trait UnicycleDrive extends Drive {
-  case class UnicycleSignal(forward: Dimensionless, turn: Dimensionless)
-  case class UnicycleVelocity(forward: Velocity, turn: AngularVelocity)
-
   type DrivetrainHardware <: UnicycleHardware
   type DrivetrainProperties <: UnicycleProperties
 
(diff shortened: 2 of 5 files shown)


kunalsheth (approved):
cool

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleMotionProfileControllers.scala
```diff
@@ -0,0 +1,89 @@
...
+      val accelerationValue = acceleration.toFeetPerSecondSquared
+
+      FeetPerSecond(
+        sqrt(
+          abs(
+            finalVelocitySquared + 2 * -accelerationValue * errorValue)))
+    }
+
+    velocityDeccel.zip(velocityAccel).zip(signError).map {
+      case ((velDec, velAcc), sign) => sign * Utility.min(
```
shadaj: Instead of using `Utility`, can you just do `velDec min crusingVelocity min velAcc`?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleMotionProfileControllers.scala
```diff
@@ -0,0 +1,89 @@
...
+
+trait UnicycleMotionProfileControllers[DriveSignal, DrivetrainHardware <: UnicycleHardware, DrivetrainProperties <: UnicycleProperties]
+  extends UnicycleCoreControllers[DriveSignal, DrivetrainHardware, DrivetrainProperties] {
+
+  /**
+    * This controller uses a control loop that results in constant acceleration,
+    * then a constant cruising velocity, and then a constant deceleration to the
+    *  target final velocity at the target position. It is meant to control
+    *  second order systems.
+    * @param initVelocity in feetPerSecond
```
shadaj: Because we no longer care about the actual value units (we only care that they are the correct quantity type), can you remove the annotations about the units of each param?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -52,53 +52,15 @@ trait UnicycleDrive extends Drive {
...
-    def closedLoopControl(unicycle: SignalLike[UnicycleSignal])
-                         (implicit hardware: DrivetrainHardware,
-                          props: DrivetrainProperties): PeriodicSignal[UnicycleSignal] = {
-      import props._
-
-      velocityControl(unicycle.map(s => UnicycleVelocity(
-        maxForwardVelocity * s.forward, maxTurnVelocity * s.turn
-      )))
+  object UnicycleControllers extends UnicycleCoreControllers[DriveSignal, DrivetrainHardware, DrivetrainProperties]
+    with UnicycleMotionProfileControllers {
```
shadaj: This needs to have the same type params as `UnicycleCoreControllers`. I'll improve this in a later PR.

shadaj (changes requested):
Almost perfect! Just a couple small changes needed (+ testing of course).

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/TwoSidedDrive.scala
```diff
@@ -49,6 +49,7 @@ trait TwoSidedDrive extends UnicycleDrive { self =>
   protected implicit val clock: Clock
   protected val updatePeriod: Time
 
+  //TODO: add docs
```
PhilipAxelrod, line 52: Oops, I don't know why that's there

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -26,13 +26,13 @@ trait UnicycleHardware {
   val turnVelocity: Signal[AngularVelocity]
 }
 
+case class UnicycleSignal(forward: Dimensionless, turn: Dimensionless)
```
PhilipAxelrod, line 29: Wait, why is this here?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
PhilipAxelrod: Btw, this doesn't compile:
Error:(56, 10) trait UnicycleMotionProfileControllers takes type parameters
    with UnicycleMotionProfileControllers {
         ^
When I have it take type parameters. I get these nice error in UnicycleDriveTest

Error:(85, 38) value UnicycleVelocity is not a member of UnicycleDriveTest.this.TestDrivetrain
      val in = Signal.constant(drive.UnicycleVelocity(MetersPerSecond(fwd), DegreesPerSecond(turn)))
                                     ^

PhilipAxelrod: By removing the drive, this fixes things. Don't know if you want things that way though

shadaj (approved):
Let's do this merge thing!

PR #19, opened 2017-01-28
Fix sporadic failures of unicycle drive tests

shadaj:
The failures were caused by a division logic to calculate the percent
error (since originally arbitrary values could be very large). This
fixes the failures from when the generated values were zero and so
the percent error would be undefined.

+ Fixes "Closed loop with only feed-forward is essentially open loop" test
+ Adds an arbitrary velocity and angular velocity generator to keep test data to sane ranges

Asana Task: https://app.asana.com/0/219272915179160/256759720407080


ghost (approved):
lgtm

PR #23, opened 2017-01-31
add SetOutput method to set output of launchpad

dtjong:
Asana Tasks: https://app.asana.com/0/219272915179160/219272365422649, https://app.asana.com/0/219272915179160/219272915179167


frc/src/main/scala/com/lynbrookrobotics/potassium/frc/SetOutput.scala
```diff
@@ -0,0 +1,17 @@
+package com.lynbrookrobotics.potassium.frc
+
+import com.lynbrookrobotics.potassium.tasks.ContinuousTask
+import edu.wpi.first.wpilibj.Joystick
+
+/**
```
shadaj, line 6: Can you replace this with a description of what the task does?

shadaj (approved):
LGTM

PR #26, opened 2017-02-01
Implement two-way signal pushing for config hot-loading

RinHa:
Asana Task: https://app.asana.com/0/219272915179160/250106024468347

### DIFF
--- a/config/src/main/scala/com/lynbrookrobotics/potassium/config/TwoWaySignal.scala
+++ b/config/src/main/scala/com/lynbrookrobotics/potassium/config/TwoWaySignal.scala
@@ -1,33 +1,64 @@
 package com.lynbrookrobotics.potassium.config
 
-trait TwoWaySignal[T] { self =>
-  def value: T
+import scala.ref.WeakReference
 
-  def reversePropagate(newValue: T): Unit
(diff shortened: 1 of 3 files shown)


shadaj:
LGTM 💯 just rebase and we'll merge.

shadaj (approved):

LordDarkula (dismissed):
nit 

line 63 Put both brackets on the same line in empty method.

config/src/main/scala/com/lynbrookrobotics/potassium/config/TwoWaySignal.scala
```diff
@@ -1,33 +1,65 @@
...
 
 object TwoWaySignal{
   def apply[T](initialValue: T): TwoWaySignal[T] = new TwoWaySignal[T] {
-    private var currentValue = initialValue
-
-    override def value: T = currentValue
-
-    override def reversePropagate(newValue: T): Unit = {
-      currentValue = newValue
+    override def handlePropogate(newValue: T): Unit = {
```
shadaj: Just putting @LordDarkula's comments inline. The bracket on the next line should just be put right after the bracket on this line so that it reads `= {}`.

PR #27, opened 2017-02-01
Finite task

LordDarkula:
Add final to methods in FiniteTask that should not be overriden.

Asana Task: https://app.asana.com/0/219272915179160/256714390946702


shadaj (approved):
LGTM

PR #29, opened 2017-02-01
Implement remaining task combinators: and, andUntilDone

shadaj:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/AndUntilDoneTask.scala
```diff
@@ -0,0 +1,31 @@
...
+  * A finite task where a continuous task is run while a finite task is running
+ *
+  * @param first the finite task to run
+  * @param second the continuous task to run at the same time
+  */
+class AndUntilDoneTask private[tasks](first: FiniteTask, second: ContinuousTask)
+  extends FiniteTask with FiniteTaskFinishedListener {
+  override def onFinished(task: FiniteTask): Unit = {
+    if (!first.isRunning) {
+      finished()
```
PhilipAxelrod, line 13: Won't this cause endless recursion? Is that the point, it keeps going until first finishes?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/ParallelFiniteTask.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+  * A finite task where two subtasks are run in parallel
+  * @param first one of the tasks to run
+  * @param second the other task to run
+  */
+class ParallelFiniteTask private[tasks] (first: FiniteTask, second: FiniteTask)
+  extends FiniteTask with FiniteTaskFinishedListener {
+  override def onFinished(task: FiniteTask): Unit = {
+    if (!first.isRunning && !second.isRunning) {
```
PhilipAxelrod, line 11: Same question as above

core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/AndUntilDoneTask.scala
shadaj: This is called when a subtask is finished, not when we are finished.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/ParallelFiniteTask.scala
shadaj: See above.

dtjong (approved):

PR #32, opened 2017-02-03
Compensate for drift

LordDarkula:
Add constant to compensate for initial drift


frc/src/main/scala/com/lynbrookrobotics/potassium/frc/TalonEncoder.scala
```diff
@@ -8,8 +8,11 @@ import squants.motion.AngularVelocity
 import squants.time.Seconds
 
 class TalonEncoder(talon: CANTalon, conversionFactor: Ratio[Angle, Dimensionless]) {
+
+  val drift: Double = talon.getPosition
```
shadaj, line 12: This isn't drift, it's just previously accumulated values.

shadaj:
Discussed, and because of potential edge cases we're not going to implement this.

PR #42, opened 2017-02-06
Make peekedController a val

shadaj:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/Component.scala
```diff
@@ -54,6 +58,10 @@ abstract class Component[T](period: Time)(implicit val clock: Clock) {
     }
     // scalastyle:on
```
PhilipAxelrod, line 59: I think you forgot to remove this

core/shared/src/main/scala/com/lynbrookrobotics/potassium/Component.scala
```diff
@@ -54,6 +58,10 @@ abstract class Component[T](period: Time)(implicit val clock: Clock) {
...
 
+    val value = currentController.currentValue(dt)
+
+    lastOutput = Some(value)
```
PhilipAxelrod: is "value" public by intention? I'd just do 
Some(currentController.currentValue(dt))

PhilipAxelrod (changes requested):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/Component.scala
shadaj: No, we want to keep it, because otherwise we have style complaints.

PhilipAxelrod (approved):
lgtm

PhilipAxelrod:
Also, you need to rebase I think

PR #54, opened 2017-02-14
Add a class to calculate the current xy position

PhilipAxelrod:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -9,6 +9,7 @@ trait UnicycleProperties {
   val maxForwardVelocity: Velocity
   val maxTurnVelocity: AngularVelocity
   val maxAcceleration: Acceleration
+  val initPosition: Point
```
shadaj: We probably don't need this, since we'll be dealing with relative motion

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
```diff
@@ -0,0 +1,39 @@
...
+  */
+class xyPositionCalculator(initPosition: Point,
+                           angle: PeriodicSignal[Angle],
+                           totalDistanceTraveled: PeriodicSignal[Length]) {
+  def this(angle: PeriodicSignal[Angle],
+           totalDistanceTraveled: PeriodicSignal[Length]) = {
+    this(new Point(Feet(0), Feet(0)), angle, totalDistanceTraveled)
+  }
+
+  private val initAngle = angle.currentValue(Seconds(0))
```
shadaj: Make angle a `Signal`, so then you just do `.get`. You shouldn't use `currentValue` in any main code, it's only available for testing purposes.

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
```diff
@@ -0,0 +1,39 @@
...
+  private val previousAngle = angle.scanLeft(initAngle)((previous, _, _) => previous)
```
shadaj: ????? what are you trying to do here?

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
```diff
@@ -0,0 +1,39 @@
...
+  private var accumPose: Point = initPosition
+
+  private val deltaDistanceTraveled = totalDistanceTraveled.scanLeft(Feet(0)){
```
shadaj: Yeah, really confused by your use of `scanLeft`. We should talk over what you're trying to do.

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
```diff
@@ -0,0 +1,39 @@
...
+    (previous, current, _) => current - previous
+  }
+
+  val xyPosition = deltaDistanceTraveled.zip(previousAngle).withCheck {
```
shadaj: Instead of using a var, this should use `scanLeft` to accumulate the pose.

shadaj (changes requested):
Have a couple of questions on your use of `scanLeft`, but otherwise it looks pretty good.

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
PhilipAxelrod: I need the previously computed value of the angle to do my computations, for which angle has to be a PeriodicSignal. In addition, converting from a PeriodicSignal that the angle will be provided with from the gyro to a Signal will be a painful

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
PhilipAxelrod: I need to get the previously computed angle

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
PhilipAxelrod: So I am trying to get a signal for the difference between the angle traveled between computations. So If we moved from 8ft to 10ft since the last computation, I want this to return 2ft

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPositionCalculator.scala
PhilipAxelrod: I am not quite sure I understand, doesn't scanLeft only use the current value and the previous value?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
PhilipAxelrod: I was thinking that having an init position would be convenient because we might have different starting configurations in auto, so if we change our initial position, we only have to change this  opposed to changing how for we drive relatively for every auto task we use.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def dot(other: Point): Area = {
+    this.x * other.x + this.y * other.y + this.z * other.z
+  }
+
+  def magnitude: Length = (x * x + y * y + z * z).squareRoot
+
+  def zip(other: Point) = Segment(this, other)
+
+  // Better solution, perhaps using type aliases?
```
PhilipAxelrod: Is there a better way to do this?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
shadaj: We still shouldn't have it in properties, because this is dependent on the auto mode chosen.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
```diff
@@ -33,6 +33,7 @@ trait UnicycleHardware {
 
   val forwardPosition: Signal[Length]
   val turnPosition: Signal[Angle]
+  val position: Signal[Point]
```
shadaj: We don't want to directly store this in hardware, but instead should construct the position signal when we need it (in a task).

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/Value3D.scala
```diff
@@ -1,11 +1,15 @@
...
+package com.lynbrookrobotics.potassium.units
 
 import squants.Quantity
 
 /**
   * Constructs a new 3D value given X, Y, and Z axes.
   */
 case class Value3D[Q <: Quantity[Q]](x: Q, y: Q, z: Q) {
+
+  def this(value3D: Value3D[Q]) = {
```
shadaj: Not sure why this is needed, this is identical to `value3d.copy()`.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
```diff
@@ -0,0 +1,58 @@
+package com.lynbrookrobotics.potassium.units
+
+import squants.space.{Area, Feet, Length}
+
+class Point(override val x: Length,
+            override val y: Length,
+            override val z: Length) extends Value3D(x, y, z) {
+  def this(value3D: Value3D[Length]) = this(value3D.x, value3D.y, value3D.z)
+  def this(x: Length, y: Length) = this(x, y, Feet(0))
+  def distanceTo(other: Point): Length = Segment(this, other).length
```
shadaj, line 10: Can you just include these and later values into Value3D?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def dot(other: Point): Area = {
+    this.x * other.x + this.y * other.y + this.z * other.z
+  }
+
+  def magnitude: Length = (x * x + y * y + z * z).squareRoot
+
+  def zip(other: Point) = Segment(this, other)
```
shadaj: Can you call this `segmentTo`?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
shadaj: Not really (unless you want to go by the `CanBuildFrom` route). If you integrate the other methods directly into `Value3D` you won't need this though.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
```diff
@@ -0,0 +1,58 @@
...
+  def containsInXY(toTest: Point): Boolean = {
+    val withinBoundaries =
+      toTest.x <= end.x &&
+        toTest.y <= end.y &&
+        toTest.x >= start.x &&
+        toTest.y >= start.y
+
+    // Uses point slope form of line to determine if the line constructed from
+    // start and end contains the given point
+    val onLine = (end.y - toTest.y) == xySlope * (end.x - toTest.x)
```
shadaj: You might want to have a tolerance here, because you are dealing with doubles.

sensors/src/main/scala/com/lynbrookrobotics/potassium/sensors/position/xyPosition.scala
```diff
@@ -0,0 +1,39 @@
+package com.lynbrookrobotics.potassium.sensors.position
+
+import com.lynbrookrobotics.potassium.units.Point
+import com.lynbrookrobotics.potassium.{PeriodicSignal, Signal}
+import squants.Angle
+import squants.space._
+import squants.time.{Milliseconds, Seconds}
+
+object xyPosition {
```
shadaj: Can you name this `XYPosition`? We want to start types with capital letters.

sensors/src/test/scala/com/lynbrookrobotics/potassium/sensors/imu/TestXYPosition.scala
```diff
@@ -0,0 +1,88 @@
...
+import squants.motion.{DegreesPerSecond, FeetPerSecond}
+import squants.space.{Degrees, Feet}
+import squants.time.Seconds
+
+class TestXYPosition extends FunSuite{
+  val origin = new Point(Feet(0), Feet(0))
+
+  test("Test moving straight at 2 ft/s for 1 sec results in moving 2 ft forward") {
+    lazy val startTime = System.currentTimeMillis
+    def timePassed = Milliseconds(System.currentTimeMillis - startTime)
```
shadaj: You really shouldn't depend on `System.currentTimeMillis` in tests, but instead should count out the number of time increments needed.

sensors/src/test/scala/com/lynbrookrobotics/potassium/sensors/imu/TestXYPosition.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    val position = xyPosition(
+      origin,
+      Signal.constant(Degrees(90)),
+      distance.toPeriodic
+    )
+
+    var i = 0
+    while (timePassed <= Seconds(1)){
+      if(i % 10000000 == 0) position.currentValue(Seconds(0.1))
```
shadaj: Yeah, not really sure what you're trying to do here. You should just be able to run `position.currentValue(Seconds(0.1))` 10 times to achieve a simulated duration of 1 second.

shadaj (dismissed):
Looks mostly good, but have a couple questions about your unit tests.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
PhilipAxelrod: Unfortunately, I can't do that because there multiplication is not defined for all possible types of Value3D. Guess we'll have to stick with this

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
PhilipAxelrod: Unfortunately I cannot because multiplication is not defined for all possible type paramaters of value3D. I can write these methods in Point because multiplication is defined for Length

sensors/src/test/scala/com/lynbrookrobotics/potassium/sensors/imu/TestXYPosition.scala
PhilipAxelrod: I need to force my periodic signal of position to constantly update for 1 second. Should I do this with the ClockMocker class we have in potassium? How would I add the ClockMocking dependency to the sensors module?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleDrive.scala
PhilipAxelrod: That would be problematic, since my class returns a PeriodicSignal of position. This  means that to get the position as a regular signal (which we will need to for initial positions), we will have to convert the position to a polling signal, which would be very convenient to do here, rather than every place that we need to get position. It also makes logging easier.

shadaj: That's why I was pushing for only using relative positions in the tasks. Hardware traits should only contain signals that are pulled directly from sensors, and should not contain intermediate calculations. So you shouldn't be converting to a polling signal, and should just directly use relative positions (and maybe take the initial position as an argument to the task).

shadaj: In general, I feel like you're overusing polling signals. I only added polling signals for the specific use case of gyros, where you want to treat a periodically updated value as a continuous stream. You really shouldn't use them whenever possible, though.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/spacialValues.scala
shadaj: OK, there might be a way to actually still make this work by doing implicit type evidence, but I'll leave that for a separate PR.

sensors/src/test/scala/com/lynbrookrobotics/potassium/sensors/imu/TestXYPosition.scala
shadaj: Whenever you call `currentValue`, you're simulating a time displacement anyway. Periodic signals don't even check `System.currentTimeMillis`, so you're just adding a delay

PhilipAxelrod: So if I just call `currentValue(Milliseconds(5))` 200 times, that would simulate this signal ticking for 1 second at a period of 5 milliseconds and 200 hertz?

RinHa (approved):
SHADAJ SAYS "IT LGTM"

PR #58, opened 2017-02-15
Fix various IMU bugs. (Physical test works, unit test fails, so removed)

nikwalia:


PhilipAxelrod (approved):
I think this could use a lot of style improvement, but we are on a tight schedule, so I am approving this.

PR #59, opened 2017-02-16
Add Funky Dashboard as implicit parameter via Dashboard Context

RinHa:


funkyDashboardContexts/jvm/src/main/scala/GroupContext.scala
```diff
@@ -0,0 +1,10 @@
+import com.lynbrookrobotics.funkydashboard.FunkyDashboard
+import com.lynbrookrobotics.potassium.Signal
+import com.lynbrookrobotics.potassium.dashboard.DashboardContext
+
+import scala.reflect.ClassTag
+
+class GroupContext(groupName: String, dashboard: FunkyDashboard) extends DashboardContext {
+  override def addDataset[T: ClassTag](name: String, dataset: Signal[T]): Unit = {
```
shadaj, line 8: You need to insert the dataset into the dashboard here.

shadaj:
I've been thinking about this, and this is going to be a lot messier than I originally thought it would. I'm going to push this off until later.

PR #62, opened 2017-02-17
Write CorrectOffsetWithLatency task for drivetrain

ghost:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreControllers.scala
```diff
@@ -75,17 +75,23 @@ trait UnicycleCoreControllers {
...
   }
 
-  def turnPositionControl(targetAbsolute: Angle)
-                         (implicit hardware: DrivetrainHardware,
-                          props: Signal[DrivetrainProperties]): (PeriodicSignal[UnicycleSignal], Signal[Angle]) = {
-    val error = hardware.turnPosition.map(targetAbsolute - _)
+  def continuousTurnPositionControl(targetAbsolute: Signal[Angle])
+    (implicit hardware: DrivetrainHardware,
+      props: Signal[DrivetrainProperties]): (PeriodicSignal[UnicycleSignal], Signal[Angle]) = {
+    val error = hardware.turnPosition.map(targetAbsolute.get - _)
```
shadaj: You should zip instead of calling `.get`.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/PeriodicSignal.scala
```diff
@@ -125,6 +125,29 @@ abstract class PeriodicSignal[+T] { self =>
     }
   }
 
+  /**
+    * Applies a timestamped fixed size sliding window over the signal
+    * @param size the size of the window
+    * @param filler the element to use to fill the window until elements are available
+    * @return
+    */
+  def slidingWithTime[U >: T](size: Int, filler: U): PeriodicSignal[Queue[(Time, U)]] = {
```
shadaj: You should probably simplify the job of this into being a `zipWithTime`, so that you can do `signal.zipWithTime.sliding(...)`.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/PeriodicSignal.scala
```diff
@@ -125,6 +125,29 @@ abstract class PeriodicSignal[+T] { self =>
...
+    // TODO: poor resolution with currentTimeMillis
```
shadaj: Agree, maybe we can introduce a way for the clock to provide the system time? And so this could take an implicit clock to read time from.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -175,4 +178,53 @@ trait UnicycleCoreTasks {
...
+    (implicit drive: Drivetrain,
+      hardware: DrivetrainHardware,
+      props: Signal[DrivetrainProperties]) extends FiniteTask {
+
+    val positionSlide = hardware.turnPosition.toPeriodic.slidingWithTime(
+      20, hardware.turnPosition.get
+    ).toPollingSignal(Milliseconds(5))(drive.clock)
+
+    override def onStart(): Unit = {
+      val targetAbsolute = timestampedOffset.map{ offsetWithTime =>
```
shadaj: The logic for control should go into a controller method, so that we can apply different variations (such as using an open-loop lower level control).

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -175,4 +178,53 @@ trait UnicycleCoreTasks {
...
+        val offset = offsetWithTime._1
```
shadaj: You can pattern match, `val (offset, time) = offsetWithTime`

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -175,4 +178,53 @@ trait UnicycleCoreTasks {
...
   }
+
+  class CorrectOffsetWithLatency(timestampedOffset: Signal[(Angle, Time)], tolerance: Angle)
+    (implicit drive: Drivetrain,
+      hardware: DrivetrainHardware,
+      props: Signal[DrivetrainProperties]) extends FiniteTask {
+
+    val positionSlide = hardware.turnPosition.toPeriodic.slidingWithTime(
+      20, hardware.turnPosition.get
+    ).toPollingSignal(Milliseconds(5))(drive.clock)
```
shadaj: If possible, don't use a polling signal and instead bring in values from this signal using `zip`. Polling signals really shouldn't be used unless you are producing sensor data, because they automatically assign a clock to the value and don't allow the component that's using the signal for control data to configure the update rate.

Also, polling signals can cause lots of leaks, because they never stop updating even if no one is using them.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -175,4 +178,53 @@ trait UnicycleCoreTasks {
...
+        val offset = offsetWithTime._1
+        val time = offsetWithTime._2
+
+        val positionHistory = positionSlide.get match {
+          case Some(positionHistoryElement) => positionHistoryElement
+          case None => Queue[(Time, Angle)]((Milliseconds(System.currentTimeMillis), hardware.turnPosition.get))
+        }
+
+        var closestTimeSoFar = positionHistory.head
+        positionHistory.foreach{ positionHistoryElement =>
```
shadaj: You can just use ```val closestTimeSoFar = positionHistory.minBy { positionHistoryElement =>
  val (positionTime, position) = positionHistoryElement
  (positionTime - time).abs.value
}```

shadaj (changes requested):
Looks good for a first start, but there are a couple of improvements that we need to make before merging.

shadaj (approved):
LGTM

PR #73, opened 2017-02-28
Add simpsons numerical integration method to periodicSignal

nikwalia:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/PeriodicSignal.scala
```diff
@@ -184,6 +184,26 @@ abstract class PeriodicSignal[+T] { self =>
...
+    * The integral of the signal's units. More appropriate for
+    * non-linear forms of motion.
+    */
+  def simpsonsIntegral[I <: Quantity[I] with TimeIntegral[_]](implicit derivEv: T => TimeDerivative[I]): PeriodicSignal[I] = {
+    // scalastyle:off
+    val previousValues = sliding(3, null.asInstanceOf[T])
+
+    previousValues.scanLeft(null.asInstanceOf[I]){(acc, velocities, dt) =>
+      if (velocities.head != null) {
+        val secondVelocity = velocities.dequeue._2.dequeue._1
```
PhilipAxelrod: Probably want to rename velocities to something else, like currentValues, or last3Values or something, because it doesn't need to be velocity

PhilipAxelrod (changes requested):
mostly lgtm, just rename 1 variable, and we're good to go

PhilipAxelrod (approved):
LGTM!

PR #74, opened 2017-03-05
Refactor implementation of pure-pursuit controller

shadaj:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
```diff
@@ -165,5 +163,70 @@ trait PurePursuitControllers {
       }
     )
   }
+
+  def followWayPointsController(wayPoints: Seq[Point], position: PeriodicSignal[Point])
+                               (implicit hardware: UnicycleHardware,
+                                signal: Signal[UnicycleProperties]): (PeriodicSignal[UnicycleSignal], Signal[Option[Length]]) = {
+    var previousLookAheadPoint: Option[Point] = None
+
+    val biSegmentPaths = wayPoints.sliding(3).map { points =>
```
PhilipAxelrod, line 172: You should prepend the initial position

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
```diff
@@ -165,5 +163,70 @@ trait PurePursuitControllers {
...
+        if (points.size == 3) {
+          Some(Segment(points(1), points(2)))
+        } else {
+          None
+        }
+      )
+    }
+
+    var currPath: (Segment, Option[Segment]) = (
+      Segment(wayPoints(0), wayPoints(1)),
```
PhilipAxelrod, line 184: Doesn't handle case of only one way point. Prepend initial position to fix this.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
```diff
@@ -165,5 +163,70 @@ trait PurePursuitControllers {
...
+    val historyUpdatingLookAheadPoint = lookAheadPoint.withCheck{p =>
+      previousLookAheadPoint = Some(p)
+    }
+
+    val forwardOutput = pointDistanceControl(
+      position,
+      historyUpdatingLookAheadPoint)._1
+    val distanceToLast = position.map(_ distanceTo wayPoints.last)
+
+    val errorToLast = distanceToLast.peek.map { d =>
```
PhilipAxelrod, line 218: This breaks  down if after traveling some path, we need to return to the starting point. Instead use the distance to the look ahead point

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitTasks.scala
```diff
@@ -0,0 +1,71 @@
...
+
+      val position = XYPosition(
+        hardware.turnPosition.map(_ - initialTurnPosition).map(compassToTrigonometric),
+        hardware.forwardPosition
+      )
+
+      val (unicycle, error) = followWayPointsController(wayPoints, position)
+
+      drive.setController(lowerLevelOpenLoop(unicycle.withCheck { _ =>
+        if (error.get.forall(_ < tolerance)) {
```
Paxelord: Why is error an option?

PhilipAxelrod (changes requested):

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
PhilipAxelrod: You'll get a indexOutOfBounds exception

PhilipAxelrod (changes requested):

PhilipAxelrod (approved):
Sure, lgtm

PR #76, opened 2017-03-07
Changed nLEDs to actual value

ghost:


ghost (approved):

shadaj:
By the way, this should be configurable, we don't want robot specific information in potassium.

PR #77, opened 2017-03-08
Fix issues where as robot approached close to target in pure pursuit …

PhilipAxelrod:
…auto, it would sharply turn 90 degrees


ghost (approved):
LGTM

PR #78, opened 2017-03-09
Use Simpson's integration method for calculating position

PhilipAxelrod:
Differentiate distance traveled from encoders to get velocity, then use simpson's intergration to get position


shadaj (approved):
LGTM

PR #81, opened 2017-03-12
Add task for turning to an absolute angle

shadaj:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -201,6 +201,26 @@ trait UnicycleCoreTasks {
...
+                      props: Signal[DrivetrainProperties]) extends FiniteTask {
+    override def onStart(): Unit = {
+      val (controller, error) = turnPositionControl(absoluteAngle)
+      val checkedController = controller.withCheck { _ =>
+        if (error.get.abs < tolerance) {
+          finished()
+        }
+      }
+
+      drive.setController(lowerLevelVelocityControl(speedControl(checkedController)))
```
PhilipAxelrod, line 216: Instead of writing a controller to convert a unicycle signal to a unicycle velocity, let's just add a constructor to unicycleVelocity that takes a unicycleSignal.

PhilipAxelrod (changes requested):
super quick change, otherwise, lgtm

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
shadaj: You would have to do a map anyway to transform the type, so that's what `speedControl` does.

PhilipAxelrod: Ok cool. Can we call this toSpeedSignal or something, because speedControl sounds like we are doing some control logic like pidf

PhilipAxelrod: I'll just do it in my next pr

PhilipAxelrod (approved):
lgtm

PR #82, opened 2017-03-16
Extrapolate look ahead points to prevent rapid turning when approaching final target way point

PhilipAxelrod:
Rather than restricting the look ahead point to be on the defined path, we instead extrapolate the line segments, and use the look ahead point on the extrapolated line. This fixes the bug where the robot turns 90 degrees upon reaching the target turns 90 degrees, because instead of turning toward the last point in the path, the robot turns toward an imaginary point ahead if itself.

### DIFF
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
@@ -3,10 +3,10 @@ package com.lynbrookrobotics.potassium.commons.drivetrain
 import com.lynbrookrobotics.potassium.control.PIDF
 import com.lynbrookrobotics.potassium.units.{Point, Segment}
 import com.lynbrookrobotics.potassium.{PeriodicSignal, Signal}
-import squants.{Angle, Dimensionless, Percent}
+import squants.{Angle, Dimensionless, Percent, Time}
 import squants.space.{Angle, _}
 
-trait PurePursuitControllers {
+trait PurePursuitControllers extends UnicycleCoreControllers {
   /**
     * use pid control on the distance from the target point
     * @param target the target point to reach
@@ -42,7 +42,8 @@ trait PurePursuitControllers {
     * @param lookAheadDistance
     * @return the look ahead point for the given path
     */
-  def getLookAheadPoint(biSegmentPath: (Segment, Option[Segment]),
+  @deprecated
+  def getNonExtrapolatedLookAheadPoint(biSegmentPath: (Segment, Option[Segment]),
                                   currPosition: Point,
                                   lookAheadDistance: Length): Point = {
     import MathUtilities._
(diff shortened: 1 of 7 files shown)


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
```diff
@@ -80,99 +81,98 @@ trait PurePursuitControllers {
   /**
     * returns an angle limited from -180 to 180 equivalent to the input
     * @param degrees
-    * @return angle betwee -270 and 90
+    * @return angle betwee -90 and 90
```
shadaj, line 84: This should be -180 to 180, right?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
```diff
@@ -42,7 +42,8 @@ trait PurePursuitControllers {
     * @param lookAheadDistance
     * @return the look ahead point for the given path
     */
-  def getLookAheadPoint(biSegmentPath: (Segment, Option[Segment]),
+  @Deprecated
```
shadaj: Use the lowercase `@deprecated`

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreControllers.scala
```diff
@@ -122,4 +122,32 @@ trait UnicycleCoreControllers {
...
+    val control = PIDF.pid(
+      hardware.turnPosition.toPeriodic,
+      targetAbsolute.toPeriodic,
+      props.map(_.turnPositionControlGains)
+    ).map(s => UnicycleSignal(Percent(0), s))
+
+    (control, error)
+  }
+
+  def turnPositionControl(targetAbsolute: PeriodicSignal[Angle])
```
shadaj: Why is this method duplicated 3 times?

shadaj (changes requested):
Couple of changes and we should be able to merge!

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreControllers.scala
PhilipAxelrod: Yeah, I'll remove these

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/PurePursuitControllers.scala
PhilipAxelrod: I deal with the case that the target angle is 180 or similar by calling this method. So if the input angle is 180, it returns 0. You are right that it is not returning an equivalent angle, but that is compensated by the fact that we drive backwards

shadaj (approved):
LGTM

PR #87, opened 2017-04-01
Add task to drive beyond a set distance

shadaj:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -183,6 +182,55 @@ trait UnicycleCoreTasks {
...
+      }
+
+      val absoluteAngle = hardware.turnPosition.get
+      val (turnController, turnError) = turnPositionControl(absoluteAngle)
+
+      val combinedController = limitedForward.zip(turnController).map(t => t._1 + t._2)
+
+      val checkedController = combinedController.withCheck { _ =>
+        val beyond = if (distance.value > 0) {
+          forwardError.get.abs.value < 0
```
PhilipAxelrod: Because you're taking the absolute value, this is always false

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/UnicycleCoreTasks.scala
```diff
@@ -183,6 +182,55 @@ trait UnicycleCoreTasks {
...
+        } else {
+          forwardError.get.abs.value > 0
```
PhilipAxelrod: See above

PhilipAxelrod (changes requested):

PhilipAxelrod (approved):

PR #88, opened 2017-04-06
Run all WPIClock periodic tasks in their own thread

shadaj:
Right now, there is no documentation on what thread WPILib runs Notifier tasks in, so this PR changes the implementation of setting up periodic tasks to use a separate thread with a semaphore transferring Notifier ticks to execution of the task within the thread.


shadaj:
Closing, because notifiers already have their own threads.

PR #92, opened 2017-05-13
Fix bugs with Simpson's integration and position calculation with Simpson's integration

PhilipAxelrod:
Fix bug where position calculations with Simpson's integration used second derivative of position instead of first. In addition, fix bug where dt was twice as much as it should be. By coincidence, the unit test used incorrect values that showed this integration method as functioning properly.

### DIFF
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/cartesianPosition/XYPosition.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/cartesianPosition/XYPosition.scala
@@ -8,40 +8,17 @@ import squants.space._
 object XYPosition {
   /**
     * returns signal of current xy position provided distance traveled since
-    * start of calculations and current angle
+    * start of calculations and current angle. It assumes that motion of the robot is
+    * a straight line between ticks, and then using vector addition to continually update
+    * the position
     * @param angle current trigonometric angle in about z axis (in the xy
-    *              plane), where 0 is defined at the x axis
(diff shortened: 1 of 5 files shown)


nikwalia (approved):
LGTM?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/cartesianPosition/XYPosition.scala
```diff
@@ -8,64 +8,68 @@ import squants.space._
...
-      Feet(0)
-    )
+      Feet(0))
 
     // approximate that robot traveled in a straight line at the average
     // angle over the course of 1 tick
     deltaDistance.zip(averageAngle).scanLeft(origin){
       case (acc, (distance, avrgAngle), _) =>
-        acc + new Point(
-          distance * Math.cos(avrgAngle.toRadians),
```
shadaj, line 38: If we're dropping use of `java.util.Math`, we should remove the import as well.

shadaj (changes requested):
Generally looks good, just one cleanup we can include in this PR

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/cartesianPosition/XYPosition.scala
PhilipAxelrod: Done

shadaj (approved):
LGTM! Thanks for the fix @Paxelord.

PR #94, opened 2017-05-15
Cross compile core, testing, control, and commons to Scala Native

shadaj:
🎉 🎉 🎉


PhilipAxelrod (approved):
lgtm. Shame that scala native doesn't support scala 2.12 though

PR #96, opened 2017-06-09
Implement Streams, the evolution of Signals

shadaj:
Things to do:
- [X] Periodic and manually pushed streams
- [X] Synchronized zipping of streams
- [x] Asynchronous zipping of streams
- [x] Eager zipping of streams
- [x] Zipping streams with time of creation
- [x] Sliding over streams
- [x] Access to delta-times when transforming streams
- [x] Scanning over streams
- [x] Integral and derivative of streams
- [x] Deferring stream values to a separate thread
- [x] Synchronize with another stream through polling
- [x] Make periodic by polling at set interval
- [x] ~~Handling stream values with constant values~~
- [x] Filtering stream values into another stream
- [x] ~~Passing a condition over a stream to create an event~~
- [x] Insert a check in a stream pipeline (withCheck)
- [x] ~~Use streams in components with verification of periodicity~~
- [x] ~~Add `Stream.detach` to manually deregister stream from parents~~
- [x] Add ability to provide custom implementations of a time source (such as a mock for `dt`s)


core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+        case None =>
+          bCancel()
+      }
+    }
+
+    ret
+  }
+
+  /**
+    * Adds a listener for elements of this Signal. Callbacks will be executed
```
PhilipAxelrod: Signal?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+    * a periodic nature.
+    * @param other the stream to merge with
+    * @tparam O the type of values from the other stream
+    * @return a stream with the values from both streams brought together
+    */
+  def zipEager[O](other: Stream[O]): Stream[(T, O)] = {
+    val ret = new EagerZippedStream[T, O]
+    val ptr = WeakReference(ret)
+
+    var aCancel: Cancel = null
```
PhilipAxelrod: Why do we have this line? Why not just var aCancel = this.foreach{ //etc?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+  /**
+    * Merges two streams, with a value published from the resulting stream
+    * whenever the parent streams have both published a new value
+    * @param other the stream to merge with
+    * @tparam O the type of values from the other stream
+    * @return a stream with the values from both streams brought together
+    */
+  def zip[O](other: Stream[O]): Stream[(T, O)] = {
+    val ret = new ZippedStream[T, O](expectedPeriodicity, other.expectedPeriodicity)
+    val ptr = WeakReference(ret)
```
PhilipAxelrod: How do we ensure the gc doesn't delete an object we need? Should we use `SoftReference` instead? Idk, I just learned about this from a blog 5 minutes ago.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+
+    var aCancel: Cancel = null
+    aCancel = this.foreach { a =>
```
PhilipAxelrod: So the idea is to update the stream one last time before canceling it?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+        case Some(s) =>
+          s.receiveA(a)
+        case None =>
+          aCancel()
+      }
+    }
+
+    var bCancel: Cancel = null
+
+    bCancel = other.foreach { b =>
```
PhilipAxelrod, line 80: Wait, how does `aCancel` and `bCancel` effect `ret`?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ExpectedPeriodicity.scala
```diff
@@ -0,0 +1,9 @@
+package com.lynbrookrobotics.potassium.streams
+
+import squants.time.Time
+
+sealed trait ExpectedPeriodicity
+
+case class Periodic(period: Time) extends ExpectedPeriodicity
```
PhilipAxelrod, line 7: Wait, what's the point of this?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+      ptr.get match {
+        case Some(s) =>
+          s.receiveFollower(b)
```
PhilipAxelrod: Wait, so Followers receive input from Masters? So why would we ever use `ReceiveMaster` ?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+    listeners.enqueue(thunk)
+
+    () => {
+      listeners.dequeueFirst(_ eq thunk)
+    }
+  }
+}
+
+object Stream {
+  def periodic[T](period: Time)(value: => T)(implicit clock: Clock): Stream[T] = {
```
PhilipAxelrod, line 415: Maybe add something like this to the `Stream` class? Sort of like `[Signal instance].toPeriodic` that we have right now?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+    new Stream[T] {
+      val expectedPeriodicity = Periodic(period)
```
PhilipAxelrod, line 417: Excuse my Scala incompetence, but why do we not have to explicitly override `expectedPeriodicity` here, but on line 159 we do?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }
+
+  def manual[T]: (Stream[T], T => Unit) = {
+    var publish: T => Unit = null
+
+    val signal = new Stream[T] {
+      override val expectedPeriodicity: ExpectedPeriodicity = NonPeriodic
+
+      publish = publishValue
```
PhilipAxelrod: Is there a difference between this and `publish = (a) => publish(a)`?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,166 @@
...
+    new Stream[T] {
+      val expectedPeriodicity = Periodic(period)
+
+      clock(period) { _ =>
+        publishValue(value)
+      }
+    }
+  }
+
+  def manual[T]: (Stream[T], T => Unit) = {
```
PhilipAxelrod, line 425: Why do you need to explicitly return a function that publishes the current value of the `Steam`? Can't you just call `[Stream instance].publishValue`?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -0,0 +1,36 @@
+package com.lynbrookrobotics.potassium.streams
+
+class ZippedStream[A, B](aPeriodicity: ExpectedPeriodicity,
+                         bPeriodicity: ExpectedPeriodicity) extends Stream[(A, B)] {
+  override val expectedPeriodicity: ExpectedPeriodicity = (aPeriodicity, bPeriodicity) match {
+    case (Periodic(a), Periodic(b)) =>
+      if (a == b) {
+        Periodic(a)
+      } else {
+        NonPeriodic
```
PhilipAxelrod: Wouldn't you want to make it at the `expectedPeriodicity` of the less frequently updating `Steam`?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -0,0 +1,36 @@
...
+      }
+
+    case _ => NonPeriodic
+  }
+
+  private[this] var aSlot: Option[A] = None
```
PhilipAxelrod, line 17: This stores the last calculated value?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/package.scala
```diff
@@ -0,0 +1,5 @@
+package com.lynbrookrobotics.potassium
+
+package object streams {
```
PhilipAxelrod, line 3: Hugh?

PhilipAxelrod (changes requested):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: The point is that we're ok with the object being deleted. If no one else is referring to the zipped stream that means it's safe to have it deleted and stop sending it updates.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Calling `foreach` adds a listener and returns a function to call when we want to cancel the subscription.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: The relationship is not follower under master, instead here master refers to the stream that determines the periodicity of the new stream, with the follower not causing the new stream to 
immediately update.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Because we refer to `aCancel` from within the callback, we need to do this awkward two-step initialization.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Oops! Still getting used to the new terminology.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Hmm, I guess that could be implemented by syncing the existing stream to a `Stream.periodic` with the given period. I'll add that to the list.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/package.scala
shadaj: I want the `Cancel` type alias to be available throughout the package, so I added it here.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
shadaj: Yep, I'm using `slot` as the term for where we hold values from parents.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
shadaj: Sometimes, even if you have a new value from the slower `Stream`, you might have to wait for a value from the newer `Stream` if it updated *just* before the slower stream updated. So this would go against matching the periodicity of the slower `Stream`.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ExpectedPeriodicity.scala
shadaj: Because `Stream`s can be transformed in ways that do not maintain their periodicity (because of the push flow), we need to keep track of if a `Stream` is periodic and if so what its period is.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Using `override` is optional because `expectedPeriodicity` is abstract. I just forgot to use it here.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: We don't want anyone with access to a `Stream` to be able to write to it, so `publishValue` is protected.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Nope, but we can keep things short by just converting the method to a function.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
PhilipAxelrod: bump.

shadaj: When they are called, they cause the parent streams to stop sending new data to `ret`, but only interact with `ret` through `ptr` to allow deallocation of the resulting stream.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
PhilipAxelrod: But the method `manual` is public anyways, so how does this change anything?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/AsyncZippedStream.scala
```diff
@@ -0,0 +1,17 @@
+package com.lynbrookrobotics.potassium.streams
+
+class AsyncZippedStream[A, B](master: ExpectedPeriodicity) extends Stream[(A, B)] {
+  override val expectedPeriodicity: ExpectedPeriodicity = master
+
+  private[this] var followerSlot: Option[B] = None
+
+  def receiveMaster(masterValue: A): Unit = {
```
PhilipAxelrod: Please rename to primary supplier and second supplier @shadaj

PhilipAxelrod: Also, please test that children carry hard references to their parents

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/AsyncZippedStream.scala
```diff
@@ -0,0 +1,17 @@
...
+    if (followerSlot.isDefined) {
+      publishValue((masterValue, followerSlot.get))
+    }
+  }
+
+  def receiveFollower(followerValue: B): Unit = {
+    followerSlot = Some(followerValue)
+  }
+}
```
kunalsheth: Does not have strong reference to parent streams?
`data.zip(otherData.map(...))`
will break stuff cause there is are no strong references to `otherData.map(...)`

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -0,0 +1,36 @@
...
+  def receiveA(aValue: A): Unit = {
+    aSlot = Some(aValue)
+    maybePublishPair()
+  }
+
+  def receiveB(bValue: B): Unit = {
+    bSlot = Some(bValue)
+    maybePublishPair()
+  }
+}
```
kunalsheth: Does not have strong reference to parent streams?
data.zip(otherData.map(...))
will break stuff cause there is are no strong references to otherData.map(...)

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/EagerZippedStream.scala
```diff
@@ -0,0 +1,24 @@
...
+  def receiveA(value: A): Unit = {
+    lastASlot = Some(value)
+    maybePublish()
+  }
+
+  def receiveB(value: B): Unit = {
+    lastBSlot = Some(value)
+    maybePublish()
+  }
+}
```
kunalsheth: Does not have strong reference to parent streams?
data.zip(otherData.map(...))
will break stuff cause there is are no strong references to otherData.map(...)

kunalsheth (changes requested):

shadaj:
@Paxelord @kunalsheth addressed your comments, can you give another review?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/EagerZippedStream.scala
```diff
@@ -0,0 +1,24 @@
+package com.lynbrookrobotics.potassium.streams
+
+class EagerZippedStream[A, B](parentA: Stream[A], parentB: Stream[B]) extends Stream[(A, B)] {
+  override val expectedPeriodicity: ExpectedPeriodicity = NonPeriodic
+
+  private[this] var lastASlot : Option[A] = None
+  private[this] var lastBSlot : Option[B] = None
+
+  def maybePublish(): Unit = {
```
ghost: This name seems a little sketch. Maybe attemptPublish?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -0,0 +1,37 @@
...
+        NonPeriodic
+      }
+
+    case _ => NonPeriodic
+  }
+
+  private[this] var aSlot: Option[A] = None
+  private[this] var bSlot: Option[B] = None
+
+  def maybePublishPair(): Unit = {
```
ghost: Same deal as other one

ghost (changes requested):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: The point is that you should only be able to manually publish values to the top-level stream created by `manual`, otherwise it would be possible to randomly publish values to intermediate streams created by transformations.

kunalsheth (approved):
Cool stuff!

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+      (q.last._1 / dt) - (q.head._1 / dt)
+    }
+  }
+
+  /**
+    * Calculates the integral of the signal, producing units of
+    * the integral of the signal's units
+    * @return a signal producing values that are the integral of the signal
+    */
+  def integral[I <: Quantity[I] with TimeIntegral[_]](implicit derivEv: T => TimeDerivative[I], clock: Clock): Stream[I] = {
```
PhilipAxelrod, line 254: I was thinking that instead of using a clock, there should be a different time stream. This way we can ensure that the current time will be constant though out the course of the calculations, regardless of how much time these calculations take

PhilipAxelrod: Wait, that's what you're doing with zipWithTime anyways. My bad.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+
+    // scalastyle:off
+    zipWithDt.scanLeft(null.asInstanceOf[I]) { case (acc, (cur, dt)) =>
+      if (acc != null) {
+        acc + (cur: TimeDerivative[I]) * dt
+      } else {
+        (cur: TimeDerivative[I]) * dt
+      }
+    }
+    // scalastyle:on
```
PhilipAxelrod, line 266: Could we add in Simpsons integration? I was hoping to use it for more accurate position tracking in pure pursuit

shadaj:
Ping @Paxelord @wafflemans re-review

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+
+abstract class Stream[T] { self =>
+  final type Value = T
+
+  val expectedPeriodicity: ExpectedPeriodicity
+
+  private[this] val listeners = mutable.Queue.empty[T => Unit]
+
+  protected def publishValue(value: T): Unit = {
+    synchronized {
```
PhilipAxelrod: Why do we need this to be synchronized? I don't imagine there'll be multiple threads sharing a common Steam and calling publishValue

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+import scala.ref.WeakReference
+
+import com.lynbrookrobotics.potassium.Platform
+
+abstract class Stream[T] { self =>
+  final type Value = T
+
+  val expectedPeriodicity: ExpectedPeriodicity
+
+  private[this] val listeners = mutable.Queue.empty[T => Unit]
```
PhilipAxelrod, line 20: Why use a queue of functions and not a queue of listening `Stream`s?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+
+    val ptr = WeakReference(ret)
+
+    var cancel: Cancel = null
+    cancel = this.foreach { v =>
+      ptr.get match {
+        case Some(s) =>
+          s.publishValue(f(v))
+        case None =>
+          cancel()
```
PhilipAxelrod: This is referencing the variable `cancel` declared on line 44?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -0,0 +1,432 @@
...
+    * whenever a new value is published in order of when the callbacks were added.
+    * Callbacks added first will be called first and callbacks added last will
+    * be called last.
+    *
+    * @param thunk the listener to be called on each published value
+    * @return a function to call to remove the listener
+    */
+  def foreach(thunk: T => Unit): Cancel = {
+    synchronized {
+      listeners.enqueue(thunk)
```
PhilipAxelrod: Can you please clarify what you mean by listeners? To me it sounds like a queue of functions to call when a new value is received, but now I'm not certain.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
PhilipAxelrod: I guess I'm not certain what this is meant to be used for

PhilipAxelrod (changes requested):
I'll finish with a more detailed review once I get back to a computer with internet. Mostly good, just some clarifications.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
PhilipAxelrod: If so can you clarify with `cancel.apply()`

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: No no, that makes a lot of sense! Because `zipWithTime` ends up giving you the current time, instead of the time when the root element was emitted. I like this idea, a root time stream, which you can use when you have some calculation that is time dependent.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: We can also add a listener using foreach, but we handle both child streams and that the same way.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Yeah, foreach takes a function to be called with each value emitted by the stream.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: I think we can add that in a separate PR.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: I think we can make this change in a separate PR though, because end users will still use zipWithTime.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: This makes sure that we're not iterating over the queue and adding elements to it at the same time.

shadaj:
@Paxelord ready for another review, I think we can push some items in your last review into a separate PR

PhilipAxelrod (approved):
A few changes for a later PR, but otherwise lgtm!

ghost (approved):
Looks good, I also have some stuff for a later PR.

ghost (approved):

PR #99, opened 2017-06-26
Bump to stable versions of dependencies and add ~/.coursier to cache

shadaj:


ghost (approved):
LGTM. A little skeptical that coursier will work considering our previous experiences with it

shadaj:
This only enables it for the Travis build, but I actually used it for the entire season this year and it worked great. The few issues that existed have now been fixed, especially as Coursier will soon become an official resolver for SBT.

PR #102, opened 2017-08-17
Add ability to relativize streams against an initial value produced by them

shadaj:
Originally brought up by @Paxelord, who needed this for pure pursuit control.


PhilipAxelrod (approved):
lgtm!

PR #104, opened 2017-09-01
Added atomic config to prevent config corruption

ghost:


config/src/test/scala/com/lynbrookrobotics/potassium/config/Test.scala
```diff
@@ -9,7 +10,11 @@ object Test extends App{
   case class RobotConfig(drive: DriveConfig)
```
shadaj: While we're poking around here, do you think it would be possible to refactor this into a proper unit test?

config/src/test/scala/com/lynbrookrobotics/potassium/config/TwoWayFileTest.scala
```diff
@@ -0,0 +1,37 @@
...
+    writer.append(initConfig)
+    writer.flush()
+
+    val configFromFile = new TwoWayFile(file).map(read[RobotConfig])(
+      (_, newValue) => write[RobotConfig](newValue)
+    )
+
+    for (_ <- 1 to 100) {
+      val newConfigValue = RobotConfig(DriveConfig(MetersPerSecond(math.random)))
+      configFromFile.value_=(newConfigValue)
```
shadaj, line 34: Instead of using _=, can you just do `value = ...`?

shadaj (approved):
LGTM, good work @wafflemans and @Paxelord 🎉

PR #105, opened 2017-09-01
Changes required for running 2017 code base with streams

shadaj:
From pair-programming with @Paxelord

### DIFF
--- a/frc/src/main/scala/com/lynbrookrobotics/potassium/frc/WPIClock.scala
+++ b/frc/src/main/scala/com/lynbrookrobotics/potassium/frc/WPIClock.scala
@@ -10,7 +10,7 @@ object WPIClock extends Clock {
     var lastTime: Option[Time] = None
     var running = false
 
-    def onTick() = {
+    val notifier = new Notifier(() => {
       if (!running) {
         running = true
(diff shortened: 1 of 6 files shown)


frc/src/main/scala/com/lynbrookrobotics/potassium/frc/WPIClock.scala
```diff
@@ -21,8 +21,14 @@ object WPIClock extends Clock {
         lastTime = Some(currentTime)
         running = false
       }
+    }
+
+    val notifier = new Notifier(() => {
+      onTick()
     })
 
+    onTick()
```
PhilipAxelrod: Is there a a reason why we're `onTick()` this before we schedule?

PhilipAxelrod (changes requested):

frc/src/main/scala/com/lynbrookrobotics/potassium/frc/WPIClock.scala
shadaj: Nope, fixed!

PhilipAxelrod (approved):
lgtm!

PR #115, opened 2017-10-09
Make left and right target velocites independent of max left and righ…

PhilipAxelrod:
…t speeds

This means that if the robot is directed to drive forward at 50% without turning, the target velocity for the left
and right sides will be the same. The difference in control is handled soley by the fact that control gains are
different for the left and right side


shadaj:
Wow good catch! I don't think this is an issue for how we tested the drivetrain though because due to corrections on both sides we ended up outputting 100% power to both sides during the test anyway.

shadaj (approved):
LGTM

PR #118, opened 2017-10-12
simplification of calculation of simpson's integral

nikwalia:


shadaj:
Do you have unit tests for simpson's integral?

nikwalia:
Yes I do, all I changed was replacing velocity calculated by derivative with a direct input of velocity (direct data instead of calculated data). So the unit tests are all still there and have been accordingly modified and work.

shadaj (approved):
LGTM!

PR #119, opened 2017-10-12
Remove PeriodicSignal and SignalLike because they are no longer used

shadaj:
Replaced by Streams 🎉


PhilipAxelrod (approved):
RIP Signals. I'll miss you `toPeriodic` ;_; 
You'll remain forever in my heart

PR #121, opened 2017-10-16
Make WPIClock exception tolerant

shadaj:


PhilipAxelrod (changes requested):
This will silence all exceptions in a component. This seems too dangerous. Maybe instead add a silenceExceptions parameter to use during competitions?

PhilipAxelrod (approved):
Lgtm

PR #122, opened 2017-10-16
Added dynamic braking

kunalsheth:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,47 @@
...
+    "||.||.||.||.", // 8
+    "|||.|||.|||.", // 9
+    "|||||.|||||.", // 10
+    "|||||||||||.", // 11
+    "||||||||||||" //  12
+  )
+    .map(_.toCharArray)
+    .map(_.map(_ == '|'))
+  private val ditherLength = ditherPatterns.map(_.length).min
+  private var tick = Byte.MinValue
```
shadaj: This may be used in different places, so instead of having a global tick you should take this as a parameter.

shadaj (changes requested):
Couple changes needed before merging.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,47 @@
...
+    "|.....|.....", // 2
+    "|...|...|...", // 3
+    "|..|..|..|..", // 4
+    "|.|.|.|..|..", // 5
+    "|.|.|.|.|.|.", // 6
+    "||.||.|.|.|.", // 7
+    "||.||.||.||.", // 8
+    "|||.|||.|||.", // 9
+    "|||||.|||||.", // 10
+    "|||||||||||.", // 11
```
PhilipAxelrod: Can we review this in person? This seems like a strange thing to do

PhilipAxelrod (changes requested):

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
shadaj: +1 This could probably just be represented by a `Byte` like in the original implementation https://github.com/Team846/code-2014/blob/7f0c8ee1bb4b1678708b78b101f46959afdb1b41/LRT14/Actuators/DriveESC.cpp#L107

kunalsheth: I was talking to Mr. G about this:
- A longer pattern length allows for more intermediate brake strengths, therefore smoother deceleration.
- Writing the patterns in hex is unnecessary. It requires documentation and is hard to visualize.  

Because [Scala doesn't have binary literals](https://stackoverflow.com/questions/7197119/how-to-write-binary-literals-in-scala), I decided to use a String to store braking patterns which are converted into Boolean arrays at startup.

shadaj: The longer pattern makes sense. Now that I think of it, we should really put together a proper algorithm for calculating the brake patterns. The patterns themselves are pretty systematic so there's no reason they need to be calculated by humans.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,41 @@
...
+    "..|..|..|..|",
+    "...|...|...|",
+    ".....|.....|",
+    "...........|"  // weakest braking
+  )
+    .map(_.toCharArray)
+    .map(_.map(_ == '|'))
+  private val ditherLength = brakingPatterns.map(_.length).min
+
+  def orBrake(max: ElectricPotential, free: AngularVelocity)
```
shadaj: This name is pretty confusing, it doesn't explain what the method is supposed to calculate. How about something like `dynamicBrakingOutput` or something else that explains the purpose and result of the method.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,41 @@
...
+             (tick: Int, target: ElectricPotential, current: AngularVelocity)
+  : Option[ElectricPotential] = {
+    val sp = target / max
```
shadaj: What is this and `pv`? Need better variable names to keep things clear.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,41 @@
...
+import squants.electro.{ElectricPotential, Volts}
+import squants.motion.AngularVelocity
+
+object DynamicBraking {
+  // | -> short
+  // . -> 0 volts
+  private val brakingPatterns = List(
+    "||||||||||||", // strongest braking
+    ".|||||||||||",
+    ".|||||.|||||",
```
shadaj, line 13: Shouldn't there be a ".||||.||||.|" or something like that after this?

kunalsheth: .||||.||||.| == .|||.|||.|||

shadaj: That doesn't seem very obvious, wouldn't having four |s in a row be different then three?

kunalsheth: No, I think they are pretty similar. What matters is the ratio of '|' to '.'s because each String represents a duty cycle.

shadaj: Okay, if it is solely based on that can you come up with an algorithm for auto generating these patterns?

PhilipAxelrod:
With the signal rectification feature on new speed controllers, we don't need this anymore.

PR #130, opened 2017-11-18
Dither braking + live patterns

shadaj:
This implements dynamic/dither braking with an algorithm that generates patterns on-the-fly instead of depending on a fixed set of patterns. This means that generated patterns are more accurate (pattern braking percentage within 0.25% of requested braking power), instead of being tied to using the closest hardcoded braking pattern.

cc @mass2010chromium who came up with the algorithm and @kunalsheth who wrote the initial implementation

### DIFF
--- a/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
+++ b/commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
@@ -3,24 +3,46 @@ package com.lynbrookrobotics.potassium.commons.electronics
 import com.lynbrookrobotics.potassium.streams.Stream
 import squants.{Dimensionless, Each, Percent}
 
-sealed trait DitherStatus
-case class NotBraking(outputPower: Dimensionless) extends DitherStatus
-case class ApplyBrakingPattern(brakingPower: Dimensionless) extends DitherStatus
+sealed trait DitherStatus {
+  def negate: DitherStatus
+}
+
+case class NotBraking(outputPower: Dimensionless) extends DitherStatus {
+  override def negate: DitherStatus = NotBraking(-outputPower)
+}
+
+case class ApplyBrakingPattern(brakingPower: Dimensionless) extends DitherStatus {
+  override def negate: DitherStatus = this
+}
 
 object DynamicBraking {
+  private def calculateBrakeAndDutyCycle(desired: Dimensionless, currentSpeed: Dimensionless): DitherStatus = {
+    if (currentSpeed < Percent(0)) {
+      calculateBrakeAndDutyCycle(-desired, -currentSpeed).negate
+    } else {
+      // speed >= 0 at this point
+      if (desired >= currentSpeed) { // going faster
+        NotBraking(desired)
+      } else { // slowing down
+        val error = desired - currentSpeed // error always <= 0
+
+        if (desired >= Percent(0)) { // braking is based on speed alone; reverse power unnecessary
+          if (currentSpeed > -error + Percent(5)) {
+            ApplyBrakingPattern(Each(-error / currentSpeed)) // speed always > 0
+          } else {
+            ApplyBrakingPattern(Percent(100))
+          }
+        } else { // input < 0, braking with reverse power
+          NotBraking(Each(error / (currentSpeed + Percent(100))))
+        }
+      }
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


mass2010chromium (changes requested):
DynamicBraking:
line 73: Simplify math expression to (accum + total) / (1 + total)
lines 77-83: Remove special-casing for <50% power
lines 64, 67: Change to a nearly-equals test or similar to avoid high total OR cap total to 20 (line 90)

shadaj:
Ping @mass2010chromium for second review

mass2010chromium (approved):
Yay?

shadaj:
Hey @kunalsheth @Paxelord can y'all review since you've been involved with implementing this?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,85 @@
...
+      calculateBrakeAndDutyCycle(-desired, -currentSpeed).negate
+    } else {
+      // speed >= 0 at this point
+      if (desired >= currentSpeed) { // going faster
+        NotBraking(desired)
+      } else { // slowing down
+        val error = desired - currentSpeed // error always <= 0
+
+        if (desired >= Percent(0)) { // braking is based on speed alone; reverse power unnecessary
+          if (currentSpeed > -error + Percent(5)) {
```
kunalsheth, line 30: I don't fully understand what this `if` statement is trying to do.

Are we saying that we want to slow down with 100% braking unless we are already close (within 5%) to the desired speed, in which case, we scale our braking power? Assuming that is the case:
-  shouldn't `Percent(5)` be configurable (a param.)? — thresholds that work on "heavy" systems would fail on "lighter" systems.
- could this be better modeled by a "breaking gain"?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
```diff
@@ -0,0 +1,85 @@
...
+        // we want full coast, so just directly output instead of building up a large total
+        (false, 0D, 0D)
+      } else {
+        // combine the accumulated error (requested power vs actual brake percentage
+        // with the braking percentage that would result from not braking next
+        val windowPattern = (accum + total) / (1 + total)
+
+        if (windowPattern >= requestedPower.toEach) {
+          // reset if overshoot -- the braking percentage even with the next value being coast was too high
+          (false, 0, (windowPattern - requestedPower.toEach) * (total + 1))
```
kunalsheth, line 77: Could we simplify `accum`?

I may be missing something, but I think the following is equivalent:
line 73: `val windowPattern = accum + (total / (total + 1))`
line 77: `(false, 0, (windowPattern - requestedPower.toEach))`

To me, the extra `(total + 1)` factor made the code harder to reason with because it obscures what `accum` really represents.

kunalsheth (changes requested):

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
mass2010chromium: Its not equivalent since total changes.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
mass2010chromium: The IF is an edge case catch: The algorithm responds slowly and inaccurately to new inputs when you pass it values too close to 100% brake.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/electronics/DynamicBraking.scala
mass2010chromium: I guess its a bit misleading, the original code for the algorithm worked exactly as you described, with ACCUM representing the total accumulated error, but was not mathematically correct because it did not take into account the differences in window sizes.

ACUUM is now the average contribution to the error per tick from the previous window(s) (hence the /(prev_total + 1), and multiplying back by (total + 1) compensates for error correctly.

kunalsheth (approved):

PhilipAxelrod:
With the signal rectification feature in the new speed controllers, we don't need this anymore.

PR #133, opened 2017-12-08
Fix bug with current limiting in negative direction

shadaj:
Also adjusts slew rate limiting to take an initial starting voltage.


PhilipAxelrod (approved):
lgtm. We do need to make sure we test this on a physical robot though to prevent future mistakes like these

PR #134, opened 2017-12-15
Stream based events

PhilipAxelrod:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -494,6 +500,20 @@ abstract class Stream[+T] { self =>
...
+  /**
+    * Returns a continuous even that is true when condition is true
+    * The pollingSource ImpulseEvent is used to check when condition is true
+    * @param condition condition of the event
+    * @param pollingSource source of when to update the check for condition
+    * @return
+    */
+  def evenWithCondition(condition: T => Boolean)
+                       (implicit pollingSource: ImpulseEvent): ContinuousEvent = {
+    new ContinuousEvent(
```
shadaj: I think we can completely eliminate pollingSources at this point and update the continuous event on each stream update.

shadaj:
You beat me to it! For implementing this, however, we should definitely move away from the `pollingSource` mechanism and have the continuous event state controlled by a listener on the stream. For something similar, you can take a look at `ImpulseEventSource` (https://github.com/Team846/potassium/blob/a9f1d226c9de585a6e2b3b027701a7d1edb06fff/core/shared/src/main/scala/com/lynbrookrobotics/potassium/events/ImpluseEvent.scala#L10) where we provide an API to create a new `ImpulseEvent` that lets us control the event and also get access to the event itself.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -494,6 +500,22 @@ abstract class Stream[+T] { self =>
...
   }
 
+  /**
+    * Returns a continuous even that is true when condition is true
+    * The pollingSource ImpulseEvent is used to check when condition is true
+    * @param condition condition of the event
+    * @param pollingSource source of when to update the check for condition
+    * @return
+    */
+  def evenWithCondition(condition: T => Boolean)
```
shadaj: Typo `even` => `event`

core/shared/src/main/scala/com/lynbrookrobotics/potassium/events/ContinuousEvent.scala
```diff
@@ -85,11 +86,130 @@ class ContinuousEvent(condition: => Boolean)(implicit polling: ImpulseEvent) {
...
   }
 
-  def unary_!(): ContinuousEvent = {
+  def unary_!(): PollingContinuousEvent = {
     Signal(!conditionValue).filter(identity)
   }
 }
+
+
+class ContinuousEvent(val condition: () => Boolean) {
```
shadaj: Hmm, now that we are externally telling the `ContinuousEvent` its state, having this `condition` parameter seems like an antipattern. Can you make `checkCondition` be more like `updateCondition`?

shadaj (changes requested):
Couple more questions, but almost there!

core/shared/src/main/scala/com/lynbrookrobotics/potassium/events/ContinuousEvent.scala
PhilipAxelrod: Is this just a question of naming, or do you mean a change in code design?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -494,6 +499,18 @@ abstract class Stream[+T] { self =>
...
 
+  /**
+    * Returns a continuous even that is true when condition is true
+    * The pollingSource ImpulseEvent is used to check when condition is true
+    * @param condition condition of the event
+    * @return
+    */
+  def eventWithCondition(condition: T => Boolean): ContinuousEvent = {
+    val (event, updateEvent) = ContinuousEvent.newEvent
+    this.foreach(_ => updateEvent.apply(lastValue.exists(condition)))
```
shadaj: The cancellation function from listening needs to be stored somewhere inside the event needs to be stored somewhere so that the listening doesn't stop after GC.

shadaj:
@Paxelord can you also update the web docs, under the `docs/` folder?

docs/src/main/tut/core/events/Continuous.md
```diff
@@ -0,0 +1,54 @@
+---
```
shadaj: Can you rename the file to lowercase `continuous` to match the others?

docs/src/main/tut/core/events/Continuous.md
```diff
@@ -0,0 +1,54 @@
...
+  override def onEnd = ...
+}
+
+robotTooFast.foreach(BlinkAlarmLEDs)
+
+```
+
+3. Run a Signal of `ContinuousTask`. 
+For use when the `ContinuousTask` is variable. When the `ContinuousEvent` starts, it aborts the current Task
+,retrieves the current `ContinuousTask` from the Signal, and runs the retrieved Task.
```
shadaj: Weird wrapping thing going on here. Probably want to put everything in one line but enable soft wraps.

docs/src/main/tut/core/events/Continuous.md
```diff
@@ -0,0 +1,54 @@
...
+
+2. Run a `ContinuousTask` when the `ContinuousEvent` is running.
+```scala
+class BlinkAlarmLEDs extends ContinuousTask {
+  override def onStart = ...
+  override def onEnd = ...
+}
+
+robotTooFast.foreach(BlinkAlarmLEDs)
+
```
shadaj: Nit: extra line

docs/src/main/tut/core/events/Continuous.md
```diff
@@ -0,0 +1,54 @@
...
+the `ContinuousEvent` as running appropriately.
+
+```scala
+val buttonPressedStream: Stream[Boolean] = ...
+val buttonPressedEvent: ContinuousEvent = buttonPressedStream.eventWhen(identity)
+
+val robotSpeedStream: Stream[Velocity] =  ...
+val robotTooFast: ContinuousEvent = robotSpeedStream.eventWhen(v > FeetPerSecond(20))
+``` 
+
```
shadaj: Nit: extra line

shadaj (changes requested):
Awesome, just a couple nits.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -19,7 +20,10 @@ abstract class Stream[+T] { self =>
 
   private[this] var listeners = Vector.empty[T => Unit]
 
+  private var lastValue: Option[T @uncheckedVariance] = None
```
shadaj: This isn't being used anymore, right?

shadaj (approved):
🚢 🚢 🚢

PR #136, opened 2017-12-18
Add a fallback method

PhilipAxelrod:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -464,6 +464,31 @@ abstract class Stream[+T] { self =>
       v
     }
   }
+
+  def withFallBack(fallBack: Stream[T @uncheckedVariance],
+                   printOnFallback: Boolean = false): Stream[T] = {
+    var defaultValueIsFresh = true
+
+    val cancel1 = fallBack.foreach(_ => defaultValueIsFresh = false)
```
shadaj, line 472: As soon as this method returns, these variables are no longer needed and may be cleaned up. You should probably create a new class for a fallback stream that stores these listeners.

This is super janky, but will be fixed with the new stream tracking mechanism, which I hope to PRify this week.

PhilipAxelrod: Wait, but they are needed though. It's to make sure that the foreach stuff is never cancelled

shadaj: I mean from the JVM perspective, those variables are not depended on by anyone and so would be cleaned out.

PhilipAxelrod: Would that mean that this line in the code for creating events

https://github.com/Team846/potassium/blob/198a1f43c570bf9ca8dc4e7a8f1a50eee685cc94/core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala#L510
 
will break?

shadaj:
@Paxelord do we want to move forward with this PR or just drop and replace with the simpler approach we discussed?

PR #137, opened 2017-12-19
Fix event stream bug

PhilipAxelrod:


PhilipAxelrod:
Sorry, re used an old branch

PR #143, opened 2018-01-12
Implement Potassium-controlled Stream cleanup that does not depend on GC

shadaj:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/unicycle/UnicycleCoreControllers.scala
```diff
@@ -98,7 +98,8 @@ trait UnicycleCoreControllers {
   def forwardPositionControl(targetAbsolute: Stream[Length])
                             (implicit hardware: DrivetrainHardware,
                              props: Signal[DrivetrainProperties]): (Stream[UnicycleSignal], Stream[Length]) = {
-    val error: Stream[Length] = targetAbsolute.minus(hardware.forwardPosition)
+    // drop one to get in sync with pid
+    val error: Stream[Length] = targetAbsolute.minus(hardware.forwardPosition).drop(1)
```
kunalsheth: What are the `drop(1)`s for?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -20,38 +16,80 @@ abstract class Stream[+T] { self =>
...
-        case None =>
-          cancel.apply()
+      override def unsubscribeFromParents(): Unit = {
+        unsubscribe.cancel(); unsubscribe = null
+      }
+
+      override def checkRelaunch(): Unit = {
+        if (!allowRestart) {
+          throw new IllegalStateException("This mapped stream has been marked as non-relaunchable. You may want to use .preserve to prevent stream shutdown.")
+        }
```
kunalsheth, line 86: It looks like this is the only place where `allowRestart` is used.
If so, what is the point of `allowRestart`?

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/unicycle/UnicycleCoreControllers.scala
PhilipAxelrod: Ditto

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/AsyncZippedStream.scala
```diff
@@ -2,7 +2,20 @@ package com.lynbrookrobotics.potassium.streams
...
+  var parentAUnsubscribe: Cancel = null
+  var parentBUnsubscribe: Cancel = null
+
+  override def subscribeToParents(): Unit = {
+    parentAUnsubscribe = parentA.foreach(this.receivePrimary)
+    parentBUnsubscribe = parentB.foreach(this.receiveSecondary)
+  }
+
+  override def unsubscribeFromParents(): Unit = {
+    parentAUnsubscribe.cancel(); parentAUnsubscribe = null
```
PhilipAxelrod, line 15: Shouldn't this the private to streams?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/EagerZippedStream.scala
```diff
@@ -1,6 +1,19 @@
...
+  var parentBUnsubscribe: Cancel = null
+
+  override def subscribeToParents(): Unit = {
+    parentAUnsubscribe = parentA.foreach(this.receiveA)
+    parentBUnsubscribe = parentB.foreach(this.receiveB)
+  }
+
+  override def unsubscribeFromParents(): Unit = {
+    parentAUnsubscribe.cancel(); parentAUnsubscribe = null
+    parentBUnsubscribe.cancel(); parentBUnsubscribe = null
```
PhilipAxelrod, line 14: See above

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/unicycle/UnicycleCoreControllers.scala
shadaj: PID controllers do not output a value for the first sensor input (because of derivative/integral). This changes the error to be in-sync with the output of the PID stream.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: A couple stream methods, like relativize and scanLeft, use map but have external history and so cannot be relaunched. This makes sure that those streams let the user know that they cannot be relaunched.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/unicycle/UnicycleCoreControllers.scala
PhilipAxelrod: Won't this already be handled by the drop(1) in PIDF? Even if it wasn't this is overkill

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -20,38 +16,80 @@ abstract class Stream[+T] { self =>
...
+  def foreach(thunk: T => Unit): Cancel = {
+    self.synchronized {
+      listeners = listeners :+ thunk
+      if (listeners.size == 1) {
+        // first listener, boot up!
+        if (hasShutdown) {
+          checkRelaunch()
+        }
+
+        subscribeToParents()
```
PhilipAxelrod, line 47: I'm not sure I get what's going on here

PhilipAxelrod (changes requested):

shadaj:
@Paxelord @kunalsheth please re-review; I dropped `drop`!

kunalsheth (approved):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/AsyncZippedStream.scala
```diff
@@ -2,7 +2,20 @@ package com.lynbrookrobotics.potassium.streams
 
 import com.lynbrookrobotics.potassium.clock.Clock
 
-class AsyncZippedStream[A, B](private val parentA: Stream[A], private val parentB: Stream[B]) extends Stream[(A, B)] {
+class AsyncZippedStream[A, B](parentA: Stream[A], parentB: Stream[B]) extends Stream[(A, B)] {
+  var parentAUnsubscribe: Cancel = null
```
PhilipAxelrod: Shouldn't this be private?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
PhilipAxelrod: Nevermind, I talked to you about this

PhilipAxelrod: Won't this subscribe to your parents multiple times?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -20,38 +16,80 @@ abstract class Stream[+T] { self =>
...
+      }
+    }
+
+    new Cancel {
+      override def cancel(): Unit = self.synchronized {
+        val beforeSize = listeners.size
+        listeners = listeners.filterNot(_ eq thunk)
+
+        if (beforeSize > 0 && listeners.isEmpty) {
+          // we are shutting down!
```
PhilipAxelrod: I think this can be replaced with just listeners.size == 0

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -20,38 +16,80 @@ abstract class Stream[+T] { self =>
...
+    * @param thunk the listener to be called on each published value
+    * @return a function to call to remove the listener
+    */
+  def foreach(thunk: T => Unit): Cancel = {
+    self.synchronized {
+      listeners = listeners :+ thunk
+      if (listeners.size == 1) {
+        // first listener, boot up!
+        if (hasShutdown) {
+          checkRelaunch()
```
PhilipAxelrod, line 44: Don't you need to set hasShutDown to false here?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -67,13 +105,13 @@ abstract class Stream[+T] { self =>
...
-    map { v =>
+    map({ v =>
       if (firstValue.isEmpty) {
         firstValue = Some(v)
       }
 
       f(firstValue.get, v)
-    }
+    }, allowRestart = false)
   }
```
PhilipAxelrod, line 114: I don't this this should be here

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -93,31 +131,7 @@ abstract class Stream[+T] { self =>
...
-      ptr.get match {
-        case Some(s) =>
-          s.receiveB(b)
-        case None =>
-          bCancel.apply()
-      }
-    }
-
-    ret
+    new ZippedStream[T, O](this, other, false)
```
PhilipAxelrod: Please use named paramater for the Boolean

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -214,37 +180,37 @@ abstract class Stream[+T] { self =>
...
+        unsubscribe = self.foreach { v =>
           if (last.size == size) {
             last = last.tail
           }
 
           last = last :+ v
 
           if (last.size == size) {
-            s.publishValue(last)
+            this.publishValue(last)
```
PhilipAxelrod, line 195: I realize this is for another pr, but we shouldn't be waiting for certain number of values before updating children. This might cause problems if we want do a large sliding window. 

Instead we should have a queue of options of just a queue that grows until it reaches size

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -508,35 +385,20 @@ abstract class Stream[+T] { self =>
...
 
 object Stream {
   def periodic[T](period: Time)(value: => T)(implicit clock: Clock): Stream[T] = {
     new Stream[T] { self =>
+      override def subscribeToParents(): Unit = {}
+      override def unsubscribeFromParents(): Unit = {}
+
       override val expectedPeriodicity = Periodic(period)
 
       override val originTimeStream = Some(new Stream[Time] {
```
PhilipAxelrod, line 397: I think this needs to be persistent

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
```diff
@@ -508,35 +385,20 @@ abstract class Stream[+T] { self =>
...
-      self.synchronized {
-        listeners = listeners.filterNot(_ eq thunk)
-      }
-    }
-  }
 }
 
 object Stream {
   def periodic[T](period: Time)(value: => T)(implicit clock: Clock): Stream[T] = {
     new Stream[T] { self =>
```
PhilipAxelrod, line 391: Doesn't this whole thing need to be persistent?

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -1,8 +1,27 @@
 package com.lynbrookrobotics.potassium.streams
 
-import com.lynbrookrobotics.potassium.clock.Clock
+import squants.time.Time
+
+class ZippedStream[A, B](parentA: Stream[A], parentB: Stream[B], skipTimestampCheck: Boolean) extends Stream[(A, B)] {
+  var parentAUnsubscribe: Cancel = null
```
PhilipAxelrod: Why null and not option

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
```diff
@@ -25,24 +44,26 @@ class ZippedStream[A, B](private val parentA: Stream[A], private val parentB: St
...
-  private[this] var bSlot: Option[B] = None
+  private[this] var aSlot: Option[(A, Time)] = None
+  private[this] var bSlot: Option[(B, Time)] = None
 
   def attemptPublish(): Unit = {
-    if (aSlot.isDefined && bSlot.isDefined) {
-      publishValue((aSlot.get, bSlot.get))
+    if (aSlot.isDefined && bSlot.isDefined &&
+      // if the streams are synced, make sure the timestamps match
+      (aSlot.get._2 == bSlot.get._2 || expectedPeriodicity == NonPeriodic)) {
```
PhilipAxelrod, line 53: This is going to cause issues when we use syncTo. For example, the time stamp when we get the camera data will never be the same as the gyro data ever

PhilipAxelrod: Also, please move the check into a separate line. Maybe define a `val` or method

PhilipAxelrod (changes requested):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: In this case, persistent doesn't make sense since there are no parents to subscribe or unsubscribe to.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: See below!

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Yeah, that could probably be introduced as a sliding variant.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: Well we're basing this off of `map` to not repeat ourselves, but here this type of resulting stream can't be restarted so we want to make sure that we notify the user.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
shadaj: In the case with the camera data, the expectedPeriodicity will be NonPeriodic, so the check is skipped!

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/ZippedStream.scala
shadaj: This is a very hot part, we don't want to allocate anything ever for sure.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: This only happens on the first listener, not every time you get subscribed to.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/streams/Stream.scala
shadaj: That only tracks if the stream has booted up at some point, not the current state of the stream.

PhilipAxelrod (approved):
Lgtm

PR #144, opened 2018-01-13
Added TalonController class

ghost:
TalonController is a wrapped version of TalonSRX. Ideally, no code should directly interact with raw TalonSRX objects. Instead, they should be controlled through TalonController methods.

TalonController also serves to set common configuration options throughout all Talon devices.


kunalsheth (approved):
We can continue adding to this as we need.

PR #146, opened 2018-01-14
Make joystick events take a stream for a source of updates rather tha…

PhilipAxelrod:
…n checking the status of every button with a new periodic stream.


kunalsheth (approved):
lgtm

PR #148, opened 2018-01-16
Don't depend on weak references for finite tasks

shadaj:


core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/FiniteTask.scala
```diff
@@ -18,14 +16,15 @@ trait FiniteTaskFinishedListener {
...
+  private var listener: Option[FiniteTaskFinishedListener] = None
 
   /**
     * Adds a task finish listener to listen to this task.
     * @param listener the listener of finish events
     */
-  def addFinishedListener(listener: FiniteTaskFinishedListener): Unit = {
-    listeners = WeakReference(listener) :: listeners
+  def setFinishedListener(newListener: FiniteTaskFinishedListener): Unit = {
+    assert(listener == None)
```
kunalsheth, line 26: Can we also print something incase the `-ea` flag is not set?

kunalsheth (dismissed):

core/shared/src/main/scala/com/lynbrookrobotics/potassium/tasks/FiniteTask.scala
shadaj: The way Scala's assert is implemented, exceptions will be thrown no matter what the flag is set to.

shadaj:
@kunalsheth ping re-review?

PhilipAxelrod (approved):
Lgtm

PR #150, opened 2018-01-20
Make getVelocity get the velocity instead of position

kunalsheth:


shadaj:
Have you gotten to test that this actually works? In my (brief) tests, I was still getting weird behavior with this change.

shadaj:
Ping @kunalsheth needs merge?

kunalsheth:
https://github.com/Team846/potassium/pull/158 takes care of this.

PR #158, opened 2018-01-29
Offload manual merge

kunalsheth:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/unicycle/UnicycleDrive.scala
```diff
@@ -2,47 +2,55 @@ package com.lynbrookrobotics.potassium.commons.drivetrain.unicycle
...
+      * Uses the childs's open loop control for the equivalent drive signal for the unicycle signal
+      *
       * @param unicycle the unicycle signal to drive with
-      * @return a signal controlled with open-loop on the parent
+      * @return a signal controlled with open-loop on the child
       */
-    def parentOpenLoop(unicycle: Stream[UnicycleSignal]): Stream[DriveSignal] = {
-      unicycle.map(convertUnicycleToDrive)
-    }
+    def childOpenLoop(unicycle: Stream[UnicycleSignal]): Stream[OpenLoopSignal] =
```
shadaj: I can't find any places where we don't follow this with a `map(openLoopToDriveSignal)`. I think we can change the return type to `Stream[DriveSignal]` and include the `openLoopToDriveSignal` transformation in the body itself.

core/shared/src/main/scala/com/lynbrookrobotics/potassium/units/Ratio.scala
```diff
@@ -20,6 +20,9 @@ case class Ratio[N <: Quantity[N], D <: Quantity[D]](num: N, den: D) {
 
   def *(value: Double): Ratio[N, D] = Ratio(value * num, den)
   def *(value: D): N = value / den * num
+  def *[D2 <: Quantity[D2]](r: Ratio[D, D2]): Ratio[N, D2] = Ratio(num, r.den) * (r.num / den)
+
+  lazy val recip: Ratio[D, N] = Ratio(den, num)
```
shadaj: I think this would probably be simpler as just a `def`, don't think we'll be saving much by making it lazy.

core/shared/src/test/scala/com/lynbrookrobotics/potassium/RatioTest.scala
```diff
@@ -1,17 +1,26 @@
 package com.lynbrookrobotics.potassium
 
 import org.scalatest.FunSuite
+import squants.Each
```
shadaj: Extra import.

shadaj (changes requested):
This is super awesome!!! Just one architecture-y idea and a two small nits, otherwise its good to go!

shadaj (approved):
Looks good!

shadaj:
@kunalsheth Just in case there are some more changes I didn't merge this. Feel free to merge this PR yourself once you feel it's ready.

kunalsheth:
@shadaj After starting work on the offloaded arm/lift, I realize that I accidentally coupled all the ESC native unit calculations with `OffloadedDrive`. I will have to refactor that logic out so it can be used in multiple places.

PR #160, opened 2018-01-30
Decouple offload logic

kunalsheth:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/offloaded/OffloadedDrive.scala
```diff
@@ -2,20 +2,20 @@ package com.lynbrookrobotics.potassium.commons.drivetrain.offloaded
...
-import squants.motion.Acceleration
 import squants.space.Length
-import squants.time.{Milliseconds, Seconds}
 import squants.{Dimensionless, Velocity}
 
 abstract class OffloadedDrive extends TwoSidedDrive {
   override type DriveSignal = TwoSided[OffloadedSignal]
   override type Properties <: OffloadedProperties
 
+  implicit val escConfig: EscConfig[Length]
```
shadaj: One thing I was thinking. Would it be possible to completely remove the translation from here and make it the responsibility of `applySignal`?

kunalsheth: @shadaj I don't see any benefit in doing that.
What purpose would `OffloadedDrive` serve then?
Didn't we decide that `applySignal` should be as low-level as possible?

kunalsheth: Didn't we decide that `applySignal` should be as low-level as possible?

shadaj: Well offloaded would still handle replacing all controllers with offloaded ones. I'm just not super comfortable with this half abstraction where normally only components care about what speed controller is used but here we're leaking that into the drive module. Just something to think about.

commons/src/main/scala/com/lynbrookrobotics/potassium/commons/drivetrain/offloaded/OffloadedProperties.scala
```diff
@@ -1,11 +1,14 @@
...
 
 import com.lynbrookrobotics.potassium.commons.drivetrain.twoSided.TwoSidedDriveProperties
+import com.lynbrookrobotics.potassium.control.offload.EscConfig
 import com.lynbrookrobotics.potassium.units.Ratio
 import squants.space.{Degrees, Length}
-import squants.{Angle, Dimensionless, Time}
+import squants.{Angle, Dimensionless}
 
 trait OffloadedProperties extends TwoSidedDriveProperties {
+  implicit val escConfig: EscConfig[Length]
```
shadaj: Hmm implicit seems a bit sketchy here since it'll only be brought into scope by a wild card import. Maybe keep it explicit?

shadaj (approved):

PR #164, opened 2018-02-03
Fix null ptr on Component init

kunalsheth:


shadaj (approved):
Yay finally we get rid of that sketchy code!

PR #166, opened 2018-02-05
Offload Lift

kunalsheth:


commons/src/main/scala/com/lynbrookrobotics/potassium/commons/lift/offloaded/OffloadedLift.scala
```diff
@@ -0,0 +1,52 @@
...
+      )
+    }
+  )
+
+  override def stayAbove(target: Stream[Length])
+                        (implicit p: Signal[Properties],
+                         hardware: Hardware): Stream[LiftSignal] = target.map { it =>
+    PositionBangBang(
+      forwardWhenBelow = true, reverseWhenAbove = false,
+      signal = p.get.toNative(it)
```
shadaj, line 37: I think this abstraction is fine, especially since in the future we may be dealing with sensors other than potentiometers. Do you think it would be possible though to have a conversion method (in the frc module) that handles converting voltage to the raw values reported by the Talon SRX?

shadaj (approved):
Looks good! Have one idea but that doesn't need to be addressed in this PR.

PR #169, opened 2018-02-07
Upickle to argonaut

kunalsheth:


build.sbt
```diff
@@ -139,11 +139,14 @@ lazy val frc = crossProject(JVMPlatform, NativePlatform).crossType(CrossType.Ful
 lazy val frcJVM = frc.jvm
 lazy val frcNative = frc.native
 
-lazy val config = project.dependsOn(coreJVM).settings(
+lazy val config = crossProject(JSPlatform, JVMPlatform, NativePlatform).crossType(CrossType.Pure).dependsOn(core).settings(
```
shadaj, line 142: You need to copy the nativeSettings call as in other cross projects here. That should fix the failing build.

kunalsheth: Hmm... That didn't seem to work.

shadaj: Oh! You've added jvmDependencies in your settings, that will need to go.

kunalsheth:
https://github.com/Team846/potassium/pull/169/commits/e907dc6798b273ad185c66de45d82cebd95b6022 was the last working version.

shadaj:
@kunalsheth the last build compiled and ran, tests failed on JS because there is no $ at the end of the class name of objects there. You should probably use `getClass.getSimpleName` in your unit tests too so that they work on all platforms.

shadaj (approved):
🚢

PR #180, opened 2018-02-20
Abstract limelight into logic into potassium and make vision angle tracking return compass angle

nikwalia:


PhilipAxelrod (approved):
Lgtm

PR #185, opened 2018-02-27
add getter for `lastControlSignal`

kunalsheth:
We may want to check the `lastControlSignal` from Component implementations to see if we accidentally output to the hardware twice within a single loop.


shadaj:
I don't think knowing what the component received will really help here. We care more that the outputted value to all speed controllers doesn't change between the end and start of consecutive ticks, which can be checked in user code.

